PR #25, opened 2019-04-23
isEditable logic was actually returning !isEditable, reversed return …

bhirt:
It appears the logic was accidentally reversed.   If you look at the example, you can only edit variables, everything else is locked.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -283,11 +283,11 @@ class MarkdownEditor extends React.Component {
       if (editor.value.activeMarks.size > 0 && editor.value.activeMarks.every(
         (mark => mark.type === 'variable'),
       )) {
-        return true;
+        return false;
       }
     }
 
-    return false;
+    return true;
   }
 
   onBeforeInput(event, editor, next) {


bhirt:
The wording on this feature was very confusing to me.   After looking more closely to the commit, I see that you actually want to only edit the variable text.   However, if you set lockText={false} in the current master, you can never edit non variable text.   So it looks like this pull request doesn't do what you intended.   However, the current master is still broken.   I can submit another pull request if you like.

jeromesimeon:
Is this reflected as an existing issue?

bhirt:
No.   I was misunderstood the lockText feature since it was broken, so the pull request should be rejected.   However, the current master is still broken.   I put a comment in the pull request.

bhirt:
This might be the intended logic:

    isEditable(editor) {
      if (this.state.lockText) {
        if (editor.value.activeMarks.size > 0 && editor.value.activeMarks.every(
          (mark => mark.type === 'variable'),
        )) {
          return true;
        }
        else {
          return false;
        }
      }
  
      return true;
    }

jeromesimeon:
I'm not that familiar enough with the code or the expected behavior to be sure. Maybe @dselman would be able to tell. I would find it useful to have an issue that explains the current behavior, the expected behavior and how this relate to the proposed fixed.

Also, if this PR is not the proposed fix, please close it and open a new one.

dselman:
> This might be the intended logic:
> 
> ```
> isEditable(editor) {
>   if (this.state.lockText) {
>     if (editor.value.activeMarks.size > 0 && editor.value.activeMarks.every(
>       (mark => mark.type === 'variable'),
>     )) {
>       return true;
>     }
>     else {
>       return false;
>     }
>   }
> 
>   return true;
> }
> ```

Thanks - I will update the logic to use this. Well spotted!

dselman:
Fixed here: https://github.com/accordproject/markdown-editor/commit/41319ba28a7c40f9a62ba4ae9a1462e71944bbdf

Issue #35, opened 2019-05-16
Reimplement formatting toolbar


dselman:
The existing floating toolbar has been disabled during the port to the latest version of React. We should implement a static formatting toolbar that can be attached to the top of the editor.

Michael-Grover:
https://projects.invisionapp.com/d/main#/console/16741009/364322244/preview
![image](https://user-images.githubusercontent.com/40569793/58038130-16d6c800-7afd-11e9-92fd-5b09e5828694.png)

assets:
[icons.zip](https://github.com/accordproject/markdown-editor/files/3198906/icons.zip)

Michael-Grover:
Updated:

https://projects.invisionapp.com/d/main#/console/16741009/364322244/preview
![image](https://user-images.githubusercontent.com/40569793/58048049-be60f400-7b17-11e9-8552-f68978b3bcc5.png)

Assets:
[icons.zip](https://github.com/accordproject/markdown-editor/files/3199618/icons.zip)

irmerk:
Next step: #20

dselman:
The vast majority of this has been completed. I'm going to close this and we will treat any remaining work as additional standalone enhancements.

PR #42, opened 2019-05-31
feat(src): implement static formatting toolbar - I35

irmerk:
# Issue #35 

### Changes
- Adjust linting
- Install icons
  - Move into svg component
  - Include specifications
- Implement static toolbar
  - Utilize portal
  - Render custom icons
  - Adjust styling
  - Optimize event handler with pointer events
- Adjust plugin icon styling
  - Color change on active state
- Create grid styling
- Remove hover component
- Adjust schema
  - Implement underline case
  - Add code support for lists
- Modularize method functionality
- List functionality
  - Handle list blocks to and from

### Flags
- ~Researching `FormatToolbar` functionality~
- ~Needs all button functionality corrected~
- ~Needs color change on active (already bolded, for example)~
- Needs parameter functionality
- Needs fix for clicking list button on a list item twice.


src/index.js
```diff
@@ -341,11 +345,13 @@ function MarkdownEditor(props) {
     const { children, mark, attributes } = props;
 
     switch (mark.type) {
+      case 'html':
```
DianaLease: why was html moved from matching the `code` case to matching the ### bold case?

src/index.js
```diff
@@ -541,6 +547,23 @@ function MarkdownEditor(props) {
     event.preventDefault();
     return false;
   });
+  /**
+   * Render the static-editing toolbar.
+   */
+  const renderEditor = useCallback((props, editor, next) => {
+    const children = next();
```
DianaLease: what does rendering `next()` as the children do here? I don't fully understand this part.

src/index.js
irmerk: This is a very good question.

jeromesimeon: I thought the answer was obvious, so that HTML would be bold?

DianaLease: but doesn't it make more sense to match the `code` style if it's html? I could be wrong though!

src/index.js
irmerk: [This](https://docs.slatejs.org/guides/rendering#the-editor-itself) would be the best I can do on a source for this.

src/index.js
irmerk: I reverted back to being with `code`

src/index.js
DianaLease: Got it, thank you!

PR #76, opened 2019-08-15
 feat(SlateAsInputEditor): forward ref to the Slate Editor

DianaLease:
### Changes
feat(SlateAsInputEditor): forward ref to the Slate Editor

This will allow us to access the slate Editor commands from the parent

### DIFF
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -15,6 +15,7 @@
 import React, {
   useEffect,
   useState,
+  useRef,
   useCallback
 }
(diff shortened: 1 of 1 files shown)


irmerk:
What are the commands we will be accessing? Trying to understand the use case.

DianaLease:
> What are the commands we will be accessing? Trying to understand the use case.

Things like `insertBlock` or `insertText` instead of rebuilding an entire new slate value when we add/edit a clause. Others may also come in handy - https://docs.slatejs.org/guides/commands-and-queries

PR #96, opened 2019-10-03
(chore) update to use markdown-transform

dselman:
Signed-off-by: Dan Selman <danscode@selman.org>

### Changes
- Update to use `markdown-transform` for Slate <> Markdown conversion
- Removed markdown <> HTML conversion from plugins

### Flags
- Plugins are now purely for controlling Slate editor behaviour, they cannot extend markdown serialisation as that is handled by markdown-transform
- Do we want to remove the `List` plugin altogether and just bake that into the `SlateAsInputEditor`?


package.json
```diff
@@ -8,7 +8,7 @@
     "url": "https://github.com/accordproject/markdown-editor.git"
   },
   "scripts": {
-    "test": "jest",
+    "test": "jest --passWithNoTests",
```
irmerk: Is this temporary until new tests are created?

src/plugins/list.js
```diff
@@ -4,53 +4,7 @@ import React from 'react';
...
-      },
-    };
-
-    const newSchema = JSON.parse(JSON.stringify(schema));
-    newSchema.blocks = { ...schema.blocks, ...additions.blocks };
-    newSchema.document.nodes[0].match.push({ type: 'ol_list' });
-    newSchema.document.nodes[0].match.push({ type: 'ul_list' });
-    return newSchema;
-  });
+  const name = 'list';
```
irmerk: Is this still used? Not seeing it in this file.

irmerk:
Curious why we remove the video plugin?

irmerk:
I think moving the `List` plugin altogether and just baking that into the `SlateAsInputEditor` makes sense, it feels like it's an integral attribute that doesn't need to be abstracted into a plugin.

package.json
dselman: Yes -- and it only seems to be needed for the Travis build. Not sure why.

src/plugins/list.js
dselman: It's used by the `PluginManager` to deterministically sort the plugins so they get applied in the same order.

dselman:
> Curious why we remove the video plugin?

Because the `markdown-transform` would need to be extended to handle the conversion from/to Slate.

Issue #98, opened 2019-10-03
Handle Underline in Markdown


irmerk:
# To Be Done Instead of OR After Issue #97

Since transitioning to [utilize the `markdown-transform`](https://github.com/accordproject/markdown-editor/releases/tag/v0.6.0), we will need to take a step back from supporting underline for the time being. Markdown doesn't natively support underline, so we need to rethink how we serialise this. 

We can then experiment with using `html` tags for underline.

HTMLGhozt:
Since #97 has a PR at I'll start to take a look here. I'm assuming we're looking at `<u></u>` to underline.

ahmedkrmn:
Since issue #97 is now closed, can I start working on this?

irmerk:
Yes @ahmedkrmn! Keep a lookout for the changes @dselman and I are making with the [`markdown-transform`](https://github.com/accordproject/markdown-transform/releases), because there may be an update that could possibly affect how you approach this. Not sure if it actually will, just keep an eye out. We should be finished with that either tomorrow or early Monday (October 6/7).

irmerk:
@ahmedkrmn what is the status of this? I'm opening it back up for the time being.

PR #99, opened 2019-10-03
chore(*): comment out underline-related code - I97

### LINKED ISSUE #97: Remove Underline Button from Toolbar
Since transitioning to [utilize the `markdown-transform`](https://github.com/accordproject/markdown-editor/releases/tag/v0.6.0), we will need to take a step back from supporting underline for the time being. Markdown doesn't natively support underline, so we need to rethink how we serialise this. 

Remove the underline option for now

We can then experiment with using `html` tags for underline.

### Related issues: 
- #81

cmccandless:
# Issue #97
Remove underline button from toolbar

### Changes
- Comment out references to underline (left as placeholders for whenever #98 is resolved)

### Related Issues
- Issue #98

Resolves #97

### DIFF
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -10,7 +10,7 @@ import * as tips from './toolbarTooltip';
 
 import * as boldIcon from '../icons/bold';
 import * as italicIcon from '../icons/italic';
-import * as underlineIcon from '../icons/underline';
+// import * as underlineIcon from '../icons/underline';
 import * as codeIcon from '../icons/code';
 import * as quoteIcon from '../icons/open-quote';
 import * as oListIcon from '../icons/OL';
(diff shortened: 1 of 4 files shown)


cmccandless:
Works just fine:
image
Bold, Italic, etc buttons still work as well.

cmccandless:
If anything, there is a slightly noticeable extra spacing created by the grid.

cmccandless:
Latest commit fixes the extra spacing:
image

irmerk:
Right, I think we also need to be editing the CSS to accommodate for this. I hope we get a solution in for handling underline soon, but in case we don't, we want the toolbar to not look like something is missing.

cmccandless:
@irmerk If there are no other changes to request, can you please mark approval?

irmerk:
You'll need to correct the DCO on your last commit.

cmccandless:
@irmerk done

Issue #101, opened 2019-10-04
Code Block Styling


irmerk:
**Is your feature request related to a problem? Please describe.**
We need to style monospace or code block text in the document

**Describe the solution you'd like**
There should be the ability to style the monospace text via a prop

**Describe alternatives you've considered**
If possible, we should support block quotes and not just monospace text.

**Additional context**
A starting point for related code is in PR #65 and #66

dkbuntzen:
Could I take this one?

PR #107, opened 2019-10-05
feat(*): styling for code blocks - I101

dkbuntzen:
# Issue #101 
Add the ability to style monospace or code block text in the document.
I am dipping my feet in the water here and feel a lot safer getting a bit of code review here before I go off on a complete tangent, feedback is most welcome.

### Changes
- Add a `codeStyle` prop to InputEditor
  - Pass this style object prop down to the `Code` styled component

### Flags
- Currently I allow the passing of an entire style object, I find this to be the most flexible but it will require a better definition of PropTypes if it is to stay. I would also possibly like to allow passing of styled-components css object or style object (but this may be overkill)
- Currently only affecting the `code` object, ideally I could restrict the styles that I allow and apply them more globally to code, pre, and blockquote elements.

### Related Issues
- Issue #101


irmerk:
@dkbuntzen just make sure to @ me when this PR is ready for review (or any questions), because I don't think I'll get notified when it's made ready.

dkbuntzen:
@irmerk I did have a few concerns with my direction, I mentioned most of them above. Just want to make sure I've got the right idea before going to far

irmerk:
Sorry for the delay, I didn't actually read what you wrote when I saw "WIP" in the title, my bad! 

I like the direction you're going so far. I added a few more people to review because this is an interesting PR that we should get more input on.

>Currently I allow the passing of an entire style object, I find this to be the most flexible but it will require a better definition of PropTypes if it is to stay. I would also possibly like to allow passing of styled-components css object or style object (but this may be overkill)

I'm not 100% sure I follow what you're saying here. I think passing in a lot for styling might not be overkill considering we may want to apply this to code + pre + blockquotes (as you mention in the next quote). Though the way you're constructing it could possibly be done differently, unsure which way is better (see below horizontal rule).

>Currently only affecting the code object, ideally I could restrict the styles that I allow and apply them more globally to code, pre, and blockquote elements.

Yes, if I understand this, I think this would be the ideal.

---

Just an idea that may spark some inspiration, you could potentially use `styled-components` slightly differently to create this code block, as shown in these three sections of `cicero-ui`:
- https://github.com/accordproject/cicero-ui/blob/master/src/Navigation/index.js#L80
- https://github.com/accordproject/cicero-ui/blob/master/src/Navigation/index.js#L29
- https://github.com/accordproject/cicero-ui/blob/master/src/Navigation/styles.js#L6

Also! As a reminder to myself as much as you, we should be updating the README with whatever changes you make to reflect new props that can be passed in to style this.

irmerk:
@dkbuntzen curious what the status of this is? Do you need any assistance?

dkbuntzen:
> @dkbuntzen curious what the status of this is? Do you need any assistance?

I am fairly happy with this. I need to fix my signoff (I have been committing with the wrong one).
Other than that I am almost ready to finalise this PR

irmerk:
Alright great. Is this going to support any style in the format one would do inline `style={{ camelCase: "something" }}`?

irmerk:
@dkbuntzen pinging you for this. Also marking this as ready.

dkbuntzen:
Sorry to be so silent, I have been flat out at work. I am happy with were this is at, unless you want me to align with block quote changes. 
I have had some trouble with the DCO thing, my git-ing is apparently not up to scratch with this king of thing

irmerk:
I think rebasing everything interactively and amending all the commits to have a signoff would be the best bet.

dkbuntzen:
> I think rebasing everything interactively and amending all the commits to have a signoff would be the best bet.

Thank you, I have done so

Issue #79, opened 2019-08-22
Support readOnly prop Dictating Toolbar Rendering


irmerk:
**Is your feature request related to a problem? Please describe.**
This component should accept an optional prop `readOnly` which toggles the toolbar rendering or not

**Describe the solution you'd like**
Conditionally render the toolbar based on the `readOnly` prop

**Describe alternatives you've considered**
N/A

**Additional context**
This relates to:
- Issue #35
- Issue #78

codingdefined:
I would like to work on it. In the current state, SlateAsInputEditor takes readOnly props but that is for Editor, should we combine it or create additional props say readOnlyToolbar to render Toolbar based on it's value?

irmerk:
Great @codingdefined! I think we should be creating an additional prop (`readOnlyToolbar`) as you said. `readOnly` will be for whether or not you can type in the editor, so it will be different.

irmerk:
Start at the work done in #112

PR #112, opened 2019-10-07
Adding readOnlyToolbar prop for toolbar rendering

codingdefined:
# Issue #79
This PR adds a new props readOnlyToolbar which will make the toolbar readonly just like the Editor.

### Changes
- Added new prop readOnlyToolbar 
  - Based on readOnlyToolbar, adding a class which will make the toolbar readonly 
  - Added a new class in style.css

### DIFF
--- a/examples/src/index.js
+++ b/examples/src/index.js
@@ -1,6 +1,6 @@
 import React, { useState, useCallback } from 'react';
 import {
-  Divider, Grid, Segment
+  Button, Divider, Grid, Segment
 } from 'semantic-ui-react';
 
 import ReactDOM from 'react-dom';
@@ -61,6 +61,7 @@ function Demo() {
    */
   const [slateValue, setSlateValue] = useState(slateTransformer.fromMarkdown(defaultMarkdown));
   const [markdown, setMarkdown] = useState(defaultMarkdown);
+  const [display, setDisplay] = useState(false);
 
   /**
    * Called when the markdown changes
@@ -81,6 +82,7 @@ function Demo() {
 
   return (
     <div>
+      
       <Segment>
     <Grid columns={2}>
       <Grid.Column>
@@ -88,7 +90,8 @@ function Demo() {
       </Grid.Column>
 
       <Grid.Column>
-      <SlateAsInputEditor readOnly={false} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
+        <Button onClick={() => setDisplay(!display)}>TOGGLE READONLY</Button>
+        <SlateAsInputEditor key={display} readOnly={display} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
       </Grid.Column>
     </Grid>
     <Divider vertical>Preview</Divider>
--- a/examples/src/index.js
+++ b/examples/src/index.js
@@ -88,7 +88,7 @@ function Demo() {
       </Grid.Column>
 
       <Grid.Column>
-      <SlateAsInputEditor readOnly={false} readOnlyToolbar={false} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
+      <SlateAsInputEditor readOnly={false} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
       </Grid.Column>
     </Grid>
     <Divider vertical>Preview</Divider>
--- a/examples/src/index.js
+++ b/examples/src/index.js
@@ -88,7 +88,7 @@ function Demo() {
       </Grid.Column>
 
       <Grid.Column>
-        <SlateAsInputEditor readOnly={false} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
+      <SlateAsInputEditor readOnly={false} readOnlyToolbar={false} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
       </Grid.Column>
     </Grid>
     <Divider vertical>Preview</Divider>
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -428,7 +428,7 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
 
   return (
     <div>
-      <ToolbarWrapper {...editorProps} id="slate-toolbar-wrapper-id" className={(props.readOnlyToolbar ? 'toolbar-readonly' : '')} />
+      { props.readOnly ? '' : <ToolbarWrapper {...editorProps} id="slate-toolbar-wrapper-id" /> }
       <EditorWrapper width={editorProps.WIDTH}>
         <Editor
           ref={editorRef}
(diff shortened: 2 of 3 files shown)


DianaLease:
> I believe the toolbar should not render at all when in read-only mode, rather than just not be clickable.

Also, it should use the same `readOnly` prop that already exists. We don't need a separate one for a read-only toolbar.

codingdefined:
@DianaLease I have done the necessary changes. Removed the readOnlyToolbar and not showing the toolbar in case of readOnly props set to true.

src/SlateAsInputEditor/index.js
```diff
@@ -428,7 +428,7 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
 
   return (
     <div>
-      <ToolbarWrapper {...editorProps} id="slate-toolbar-wrapper-id" />
+      { props.readOnly ? '' : <ToolbarWrapper {...editorProps} id="slate-toolbar-wrapper-id" /> }
```
irmerk: Could this be changed to `{ !props.readOnly && <ToolbarWrapper ...`?

codingdefined: @irmerk It's done, please have a look

irmerk:
@DianaLease if you have a chance to look this over. I'll have more time to on Monday if not.

src/styles.css
```diff
@@ -31,4 +31,8 @@ body {
     display: grid;
     grid-template-columns: 3% auto 3%;
     grid-template-rows: 100%;
+}
+
+.toolbar-readonly {
```
DianaLease: this can now be removed now that we are conditionally rendering the toolbar

codingdefined: @DianaLease It's done, remove the unused CSS

irmerk:
Testing this locally, the toolbar is rendering if `readOnly` is true or false. Unsure what is going wrong here...

codingdefined:
@irmerk It works for me, when the readonly is true
image

irmerk:
Ah, got it to work! Although, we want this to work dynamically while running. I set up a little button in your branch's demo: 

```js
...

import {
  Button, Divider, Grid, Segment
} from 'semantic-ui-react';

...

function Demo() {
  /**
   * Current Slate Value
   */
  const [slateValue, setSlateValue] = useState(slateTransformer.fromMarkdown(defaultMarkdown));
  const [markdown, setMarkdown] = useState(defaultMarkdown);

  const [display, setDisplay] = useState(false);
 
 /**
   * Called when the markdown changes
   */
  const onMarkdownChange = useCallback((markdown) => {
    localStorage.setItem('markdown-editor', markdown);
  }, []);

  /**
   * Called when the Slate Value changes
   */
  const onSlateValueChange = useCallback((slateValue) => {
    localStorage.setItem('slate-editor-value', slateValue.toJSON());
    const markdown = slateTransformer.toMarkdown(slateValue);
    setSlateValue(slateValue);
    setMarkdown(markdown);
  }, []);

  return (
    <div>
      <Button onClick={() => setDisplay(!display)}>TOGGLE READONLY</Button>
      <Segment>
    <Grid columns={2}>
      <Grid.Column>
        <MarkdownAsInputEditor readOnly={true} plugins={plugins} markdown={markdown} onChange={onMarkdownChange}/>
      </Grid.Column>

      <Grid.Column>
      <SlateAsInputEditor readOnly={display} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
      </Grid.Column>
    </Grid>
    <Divider vertical>Preview</Divider>
  </Segment>
    </div>
  );
}
```

This is rendering, but when I click it and it no longer renders, clicking it a second time renders it but invisible.

codingdefined:
@irmerk Should I make this change in the current PR?

irmerk:
@codingdefined yes I think we want to have this work during runtime, to support the ability to toggle it on or off by the user in the future.

codingdefined:
@irmerk Its weird, it's not showing the toolbar on toggle but when you click inside the editor it start showing up. Not sure what is happening.

irmerk:
> Edit: Fixed it by sending the new key i.e. {display} if there is a toggle

Can't say I understand why this is the case. It clearly works in the Netlify deployment. Great work! Want to get feedback from @dselman first before merging.

examples/src/index.js
```diff
@@ -81,14 +82,16 @@ function Demo() {
...
       <Segment>
     <Grid columns={2}>
       <Grid.Column>
         <MarkdownAsInputEditor readOnly={true} plugins={plugins} markdown={markdown} onChange={onMarkdownChange}/>
       </Grid.Column>
 
       <Grid.Column>
-        <SlateAsInputEditor readOnly={false} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
+        <Button onClick={() => setDisplay(!display)}>TOGGLE READONLY</Button>
+        <SlateAsInputEditor key={display} readOnly={display} lockText={true} plugins={plugins} value={slateValue} onChange={onSlateValueChange} editorProps={propsObj} />
```
dselman: I don't understand why we need both `key` and `readOnly` if they have the same value.

DianaLease: I am also confused by this. A `key` is usually required when creating a list of React elements, where each element has a unique `key` attribute (https://reactjs.org/docs/lists-and-keys.html). I do not understand the way it is being used here.

codingdefined: @dselman & @DianaLease A `Key` need not have to be `{display}` but it can be `Math.random()` also, the problem is if we do not give a key which changes on the button click, the child component's toolbar does not rerender. So, a key should be provided which can be `display` (Changes true and false on a button click), `Math.Random()` (generates a random number every time button is click) or even `new Date()` (generates new date every time a button is clicked). 

To rerender the child component we need to have a key which changes on button click @irmerk.

DianaLease: @codingdefined This feels a bit hacky to me. The last sentence of the React docs [here](https://reactjs.org/docs/reconciliation.html) explains, "Keys should be stable, predictable, and unique. Unstable keys (like those produced by Math.random()) will cause many component instances and DOM nodes to be unnecessarily recreated, which can cause performance degradation and lost state in child components."

When the `readOnly` prop changes, it will cause a re-render of the `SlateAsInputEditor` with the new prop. It's possible though that Slate is doing some optimization at the editor level, which we have seen happen in other places. I would try checking the readOnly prop on [this line](https://github.com/accordproject/markdown-editor/blob/f5105bc2d760a0cf70c98f73e8287bb51089586e/src/SlateAsInputEditor/index.js#L415) to conditionally render the Toolbar. The other option I can think of is also passing readOnly as a prop to ToolbarWrapper [here](https://github.com/accordproject/markdown-editor/blob/f5105bc2d760a0cf70c98f73e8287bb51089586e/src/SlateAsInputEditor/index.js#L431) and using the prop in the styled component to toggle a display: none property.

irmerk:
@codingdefined is this still a work in progress?

examples/src/index.js
codingdefined: @DianaLease None of the two options actually solved the issue. 

And I guess, the keys in this sense is predictable because we are setting it as true or false, though its a hack it might not cause any performance degradation.
cc : @irmerk

irmerk:
Leaving this open until we hear from @DianaLease and @dselman

irmerk:
Closing as stale, referencing in the issue.

PR #92, opened 2019-10-01
feat: add serif `font-family` for all headings

RanadeepPolavarapu:
# Issue #22 
<OVERALL SUMMARY OF PULL REQUEST>

### Changes
- added a serif font family for all headings including the toolbar. 

### Flags
- <DESCRIBE ISSUES OR HOLDS FOR REVIEWERS>

### Related Issues
- Issue #<NUMBER HERE>
- Pull Request #<NUMBER HERE>

Issue #16, opened 2019-02-23
Support read-only zones in the editor


dselman:
Define a mark that locks selective text in the editor.

jeromesimeon:
Should that be a global flag? "Lock text" option?

dselman:
Slate already has a global flag. We ended read-only sections, possibly via a plugin.

dselman:
This has been implemented via the VariablePlugin.

PR #114, opened 2019-10-07
refactor(SlateAsInputEditor): use lockText prop for editable boolean - I16

jolanglinais:
# Issue #16
Update proposal for how we handle the lockText. 

### Changes
- Replace [this check](https://github.com/accordproject/markdown-editor/blob/master/src/SlateAsInputEditor/index.js#L278) with `if(lockText) {...` 
- Slight refactoring of the new `Heading` code from a previous PR

### Flags
- Though it would remove quite a bit of code that I’m not sure fundamentally why is there in the first place.
  - @dselman need feedback on what was meant by [this comment](https://github.com/accordproject/markdown-editor/blob/master/src/SlateAsInputEditor/index.js#L127)

### Related Issues
- Pull Request #92


irmerk:
This does not work dynamically at runtime, Slate does not accept `lockText` changing.

src/SlateAsInputEditor/index.js
```diff
@@ -190,13 +235,13 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
       case 'block_quote':
         return <blockquote {...attributes}>{children}</blockquote>;
       case 'code_block':
-        return <pre {...attributes}>{children}</pre>;
```
dselman: I don't understand this. Shouldn't we be letting Slate control the `contentEditable`?

Issue #21, opened 2019-02-23
Enter after a video duplicates it


dselman:
Pressing enter after a video tag duplicates it.

pramodrhegde:
Please assign this to me.

pramodrhegde:
@dselman @irmerk Raised PR here #116 Please review.

irmerk:
The video plugin was removed.

PR #116, opened 2019-10-08
fix(plugins): stop video duplication on enter - I21

pramodrhegde:
# Issue #21
Fix for video duplication on Enter keypress.

Cause: video duplication occurs if you are pressing enter when the current focus is on the video element.

### Changes
- checking the currently focused element in `onEnter` event. if it is not a video element then only call the `next` method.

### Flags
- @dselman Please do let me know if there is a better approach than this one to solve the issue.

### Related Issues
- Issue #21


irmerk:
@pramodrhegde same as the other comment, definitely will get to reviewing this, just working on another part of `markdown-editor` and want to get that through before I come back to review this.

irmerk:
Can you run `npm audit`? You may need to run an `npm audit fix` for vulnerabilities.

This works for me locally and seems sound. @dselman thoughts?

pramodrhegde:
@irmerk Thanks for the review! I have made the changes you have requested.

irmerk:
@pramodrhegde this may be hacky, but can you try making a brand new branch off of this one, make sure everything is installed and updated with `npm`, and then make a new PR?

pramodrhegde:
@irmerk Sure, let me try that.

pramodrhegde:
@irmerk Tried creating a fresh PR #139, the build is failing.

Just to make sure, is `js-experiment-markdown` is the right base branch I am working on? as video plugin is not yet available in master.

Also noticed that npm scripts differ from master to `js-experiment-markdown`

irmerk:
AH! That's it! I never looked at the branch the PR was aimed at. No, we'd want to be merging to master. This may be a mistake because @dselman had removed the video plugin (I think), so I'm hoping for his input here. This may be a case of the issue being outdated and I should have closed it sooner.

I pinged @dselman so we should hear from him soon.

PR #117, opened 2019-10-08
feat(SlateAsInputEditor): pass props down to slate Editor

DianaLease:

### Changes
- feat(SlateAsInputEditor): pass props down to slate Editor
- this way, any props passed to SlateAsInputEditor will also get passed to the Editor from Slate


DianaLease:
> markdown-editor must remain independent of Clauses and cicero-ui...

`clausePluginProps` is optional here, so I think it's okay. Or is there more to it than that?

Issue #2, opened 2019-01-07
Unit tests


dselman:
We need unit tests for the JSON <-> MD conversion

jeromesimeon:
Any thought about how those would be written? Using jest or more traditional mocha-style tests?

jeromesimeon:
Roundtrip tests would be useful. To answer my own question, jest seems like a fine choice. Here is a small example:

```
test('can roundtrip basic text', () => {
  const markdownText = 'This is some text.';
  const value = fromMarkdown.convert(markdownText);
  const markdownRound = toMarkdown.convert(value);
  expect(markdownRound).toEqual(markdownText);
});
```

Note that even that first test fails round-tripping:
```
  ● can roundtrip basic text

    expect(received).toEqual(expected) // deep equality

    Expected: "This is some text."
    Received: "This is some text.

    "
```

This may tie into some of the whitespace issues mentioned in https://github.com/accordproject/cicero-ui/issues/43

irmerk:
### Hacktoberfest:
This issue can provide for multiple pull requests.

codingdefined:
@irmerk I would like to work on it. Does the tests should be only on the JSON <-> MD conversion? What all scenarios should we consider for the testing, like all the buttons in the toolbar?

irmerk:
@codingdefined this would be great to pick up! Just keep aware of the changes we're making in PRs #114 and #117 over the next few days. 

We're using the [`markdown-transform`](https://github.com/accordproject/markdown-transform) now, so the conversions are described more in that README. I think the more tests the better, essentially. Testing buttons, testing copy-pasting, testing adding a clause template... If you'd use the editor in X way, we should test X.

codingdefined:
@irmerk Created the pull request for the same.

irmerk:
As mentioned in #134, this issue can provide for many PRs, so I will open this up.

irmerk:
Please note the conversation in [`markdown-transform` #117](https://github.com/accordproject/markdown-transform/issues/117) for testing this repo.

irmerk:
Closing for https://github.com/accordproject/web-components/issues/14

PR #134, opened 2019-10-18
test(SlateAsInputEditor): init unit tests for md output - I2

codingdefined:
# Issue https://github.com/accordproject/markdown-editor/issues/2
Added Unit tests for the round trip conversion from normal text to markdown.

### Changes
- Added Enzyme and Enzyme Adapter package
- Added test file with the default value and checking with expected and received.

Let me know if we need to add more tests.

image

### DIFF
--- a/src/SlateAsInputEditor/index.test.js
+++ b/src/SlateAsInputEditor/index.test.js
@@ -1,0 +1,62 @@
+import React from "react";
+import Enzyme, { shallow } from "enzyme";
+import SlateAsInputEditor from "./index";
+import Adapter from "enzyme-adapter-react-16";
+
+import NoEdit from '../../src/plugins/noedit';
+import List from '../../src/plugins/list';
+import { SlateTransformer } from '@accordproject/markdown-slate';
+
+const plugins = [NoEdit(), List()];
+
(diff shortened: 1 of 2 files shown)


src/SlateAsInputEditor/index.test.js
```diff
@@ -0,0 +1,62 @@
...
+
+const slateTransformer = new SlateTransformer();
+const propsObj = { width: '600px'};
+
+const onSlateValueChange = () => {
+    // A dummy onChange props
+  };
+
+describe("SlateAsInputEditor component", () => {
+    test("component mount", () => {
```
irmerk: I think this test could be split into two maybe. There's the `component mounts` which is line 59-60. Then there is a `successful roundtrip` which is line 54-57.

irmerk:
> Let me know if we need to add more tests

I'd say no for this PR. This is a great start, I think #2 can provide for multiple PRs.

src/SlateAsInputEditor/index.test.js
codingdefined: @irmerk Done, I have split the test cases into 2

Issue #23, opened 2019-02-23
Optionally hide the markdown tab/pane


dselman:
Some users will want to limit the users to only editing using the rich text pane.

jeromesimeon:
An options menu might be nice (e.g., markdown on flag, show compiled ergo flag)?

dselman:
I will add a `props` so that the "Edit" toggle can be hidden.

johnnyqbui:
is this one still open? I'd like to take this.

irmerk:
@dselman I believe this issue may be out of date? It sounds like it relates to #79 a bit.

irmerk:
I think go for it @johnnyqbui if you still want to. It seems like this does in fact relate to #79 and this comment I made here https://github.com/accordproject/markdown-editor/pull/112#issuecomment-542854144

PR #145, opened 2019-10-23
Add option to hide markdown pane

johnnyqbui:
Signed-off-by: J <johnnyqbui7@gmail.com>

# Issue #23 
Add option to toggle display of markdown editor.

### Changes
- Add `display` prop to `MarkdownAsInputEditor`
  - add `display PropTypes`
- Cleaned up code
- Add `display` prop which takes a boolean to toggle display of markdown editor.

### Flags
- <DESCRIBE ISSUES OR HOLDS FOR REVIEWERS>

### Related Issues
- Issue #79 
- Issue #112

### DIFF
--- a/examples/src/index.js
+++ b/examples/src/index.js
@@ -1,6 +1,6 @@
 import React, { useState, useCallback } from 'react';
 import {
-  Divider, Grid, Segment
+  Button, Divider, Grid, Segment
 } from 'semantic-ui-react';
 
 import ReactDOM from 'react-dom';
@@ -61,6 +61,7 @@ function Demo() {
    */
   const [slateValue, setSlateValue] = useState(slateTransformer.fromMarkdown(defaultMarkdown));
   const [markdown, setMarkdown] = useState(defaultMarkdown);
+  const [displayMarkdownPane, setDisplayMarkdownPane] = useState(true);
 
   /**
    * Called when the markdown changes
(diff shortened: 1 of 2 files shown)


irmerk:
@dselman and @DianaLease, I was thinking issue #23 was more about functionality, not necessarily about just the demo having this?

DianaLease:
> @dselman and @DianaLease, I was thinking issue #23 was more about functionality, not necessarily about just the demo having this?

We don't have a single component that includes both panes. The demo is loading both panes as separate components, so I think the demo is the only place this functionality makes sense, unless I am misunderstanding something.

irmerk:
> We don't have a single component that includes both panes. The demo is loading both panes as separate components, so I think the demo is the only place this functionality makes sense, unless I am misunderstanding something.

I suspect the issue is just outdated, I remember early on there being an idea of having the capability of switching between the markdown and the rich text?

DianaLease:
> I suspect the issue is just outdated, I remember early on there being an idea of having the capability of switching between the markdown and the rich text?

Yes, I believe this issue is no longer relevant.

irmerk:
I'm closing the issue and this PR. Apologies @johnnyqbui, but you're already getting all the swag for Hacktoberfest so hopefully this isn't too frustrating.

johnnyqbui:
@irmerk ok no worries, ill look for another issue to work on. Thanks!

Issue #122, opened 2019-10-16
Toolbar Buttons Work in lockText Mode


irmerk:
**Describe the bug**
Text formatting is currently being allowed during `lockText`

**To Reproduce**
1. Run locally within `cicero-ui`
  a. Download both `markdown-editor` and `cicero-ui`
  b. Run in `markdown-editor`
    - `npm install`
    - `npm run build`
    - `npm run transpile`
    - `npm link`
  c. Run in `cicero-ui`:
    - `npm install`
    - `rm -rf node_modules/\@accordproject/markdown-editor`
    - `npm link @accordproject/markdown-editor`
    - `npm run start`
2. Go to `http://localhost:8080/`
3. Ensure the `lockText` prop in `cicero-ui` being passed down to the `ContractEditor` is true.
4. Highlight text within a clause (not a variable), and try bolding or italicizing text.

**Expected behavior**
`lockText` should prevent any edits to text within a clause - except for variables

**Additional context**
N/A

Ian-Yy:
Hi @irmerk , Can I try working on this issue? I'm still quite beginner. So, sorry if I made mistake 🙏

irmerk:
@Ian-Yy any progress on this? Opening back up in the meantime.

Ian-Yy:
@irmerk I'm still trying to figure out the link between cicero-ui and markdown-editor.

irmerk:
> @irmerk I'm still trying to figure out the link between cicero-ui and markdown-editor.

Do you mean `npm link` for development? Or just the connection between the two?

Ian-Yy:
Just the connection between the two. I can see when I toggle LockText, the state is changing true and false but thinking how the state is connected to markdown-editor

Ian-Yy:
@irmerk And another thing to confirm. Do I need to keep running `npm link` everytime I change something in markdown-editor?

irmerk:
@Ian-Yy once `markdown-editor` is linked, you'll just need to run `npm run transpile` in it to refresh the compiled build. 

I think the `lockText` value is passed to `markdown-editor` (`ContractEditor`), so that should be how they are connected.

Ian-Yy:
> @Ian-Yy once `markdown-editor` is linked, you'll just need to run `npm run transpile` in it to refresh the compiled build.

@irmerk Thank you! This is what I'm looking for. Now I can continue debugging.

Ian-Yy:
[...]

PR #150, opened 2019-10-24
fix(FormattingToolbar): toolbar buttons work in locktext mode

Ian-Yy:
# Issue #122 
Currently, formatting tools such as Bold, Italic is working although lockText is true.

### Changes
Added checking on Editable before calling toggleMark

### DIFF
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -110,9 +110,13 @@ export default class FormatToolbar extends React.Component {
    * When a mark button is clicked, toggle the current mark.
    */
   onClickMark(event, type) {
-    const { editor } = this.props;
-    event.preventDefault();
-    editor.toggleMark(type);
+    const { editor, pluginManager } = this.props;
+    const { value } = editor;
+
+    if(pluginManager.isEditable(editor.value,type)) {
+      event.preventDefault();
+      editor.toggleMark(type);      
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


irmerk:
I'm testing this locally and it looks like all the buttons work within a clause template.

Ian-Yy:
Is it? I tested fine earlier. Let me retest it again

Ian-Yy:
> I'm testing this locally and it looks like all the buttons work within a clause template.

I can't seem to simulate this. Bold, Italic button doesn't work when I'm in the clause template
But I found another issue that it's not reacting to lockText. Will fix this.

Ian-Yy:
@irmerk I'm sorry but can you try again?

irmerk:
Still isn't working for me. Loom recording here: https://www.loom.com/share/075f3393ebe44af79bde0ebb8a078c9f

@DianaLease could you try this locally?

DianaLease:
> Still isn't working for me. Loom recording here: https://www.loom.com/share/075f3393ebe44af79bde0ebb8a078c9f
> 
> @DianaLease could you try this locally?

Running this locally and using npm link to use it in cicero-ui, I see the correct behavior. I cannot apply styles to locked text within a clause.

DianaLease:
> But I found another issue that it's not reacting to lockText. Will fix this.

This seems to work for me. `lockText` only affects the text within a clause. So when lockText is true, you should not be able to style locked text within a clause, but you should still be able to style other contract text, which is what is happening in this PR.

Looks good!

Ian-Yy:
Thank you for checking it and merging it up! 😄

Issue #143, opened 2019-10-22
Inline Hyperlink Tooltip


irmerk:
**Is your feature request related to a problem? Please describe.**
The hyperlink tooltip should stem from the highlighted text inline which is being selected to convert into a hyperlink.

**Describe the solution you'd like**
Would it be possible for this UI to be placed above or below the text, so that it is easier for the user to tell which text is becoming a Hyperlink?

**Describe alternatives you've considered**
This will require investigation of [SlateJS](https://docs.slatejs.org/)

**Additional context**
Related Issues:
- #100

Screenshots:

![Screen Shot 2019-10-03 at 15 17 51 PM](https://user-images.githubusercontent.com/36460856/66157152-26a90180-e5f1-11e9-90f2-d3da944c8750.png)

![image](https://user-images.githubusercontent.com/40569793/67298295-0d40fa00-f4b9-11e9-86f4-3e669eab72f8.png)

husseyexplores:
I will look into this and hopefully will open a PR regarding this shortly. I've researched the crux of the problem i.e getting the selected text position, will try to implement and see where things go.

PR #142, opened 2019-10-21
feat(FormattingToolbar): implement hyperlink tooltip - I100

sahalsaad:
# Issue #100
Replace js alert when setting the link to use popup 

### Changes
alt text

### Flags
- Not sure why the selection color gone once I select the text input.

Issue #100, opened 2019-10-03
Hyperlink Tooltip


irmerk:
**Is your feature request related to a problem? Please describe.**
We need to style hyperlink text in the document, as well as create a popup tooltip for setting the link.

**Describe the solution you'd like**
There should be a popup tooltip _similar to_ the one used in Google Docs. Use this as a vague reference, we are not looking to exactly copy it.

![Screen Shot 2019-10-03 at 15 17 51 PM](https://user-images.githubusercontent.com/36460856/66157152-26a90180-e5f1-11e9-90f2-d3da944c8750.png)

**Describe alternatives you've considered**
N/A

**Additional context**
The tooltip that pops up for buttons in the toolbar currently were implemented in PR #66

ahmedkrmn:
Can I work on this?

ahmedkrmn:
Hi @irmerk, just a quick question to make sure I'm on the right track.
Currently the link pop up is a JavaScript `alert` that looks as follows:

_Add Link:_

![link](https://user-images.githubusercontent.com/20889958/66259192-62d59100-e7ae-11e9-8c05-6487f8ca5a59.PNG)
_Add Text:_


![text](https://user-images.githubusercontent.com/20889958/66259194-6537eb00-e7ae-11e9-945e-6212e333f0f6.PNG)

---

What you mean by **"Hyperlink Tooltip"** is that you want replace the two alerts with a single popup that looks similar to the one used in Google Docs?

irmerk:
@ahmedkrmn yes, that's what I was thinking. And again, doesn't need to be _specifically_ like Google Docs, that's just an example. But definitely moving away from the alert style and more of a popup with text fields.

ahmedkrmn:
@irmerk Can I use an `npm` React component or a JavaScript package, or should the new code be entirely written from scratch in JavaScript/CSS ?

irmerk:
I'd hesitate to implement new libraries, especially if it's for this one component. We are utilizing `styled-components` and `CSS Grid` so far, though.

irmerk:
This is open again. Any other contributor should view #108 as a starting point or for some context.

sahalsaad:
Hi! I'll work on this issue. Thanks

PR #157, opened 2019-10-25
Husseyexplores issue 143 link tooltip pos

husseyexplores:
# Issue #143 
Positions the hyperlink popup inline, relative to the selected text.

### Changes
- Gets the selected text node via DOM `getSelection()` api then calculates the position of the node in the DOM using hyperlink popup and selected nodes' `x, y, width height`.
- Replaces the `createElement` function with `styled.div` and uses `as` prop in `SlateAsInputEditor`. The reason for doing that is the selected text was getting unselected on all elements once we click the hyperlink toggle button.

Reference from PR #142
>Not sure why the selection color gone once I select the text input.

Above change basically, fixes that. `as` prop is supported by styled-components so it also makes the code a bit simpler. Now the selected text remains selected even upon clicking the hyperlink toggle button.
Gif

### Flags
\-

### Related Issues
- Issue #100 
- Pull Request #142

### DIFF
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -1,5 +1,5 @@
 import React, { createRef } from 'react';
-import ReactDOM from 'react-dom';
+import ReactDOM, { findDOMNode, createPortal } from 'react-dom';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import {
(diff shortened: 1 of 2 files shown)


husseyexplores:
What is this DCO error? 😥 It is asking me to run the `--signoff` command.
I've tried it but it still shows the same message. 
I've also tried rebasing (IDK if it even worked). Should I just open a new PR with fresh commits signed from the beginning?

irmerk:
Yeah that is a good plan B. If you interactive rebase, you should be able to amend each commit with a signoff.

PR #158, opened 2019-10-25
refactor(SlateAsInputEditor): remove styling props so user can style with own css

DianaLease:
refactor(SlateAsInputEditor): remove styling props so user can style with own css

# Issue
- Passing in props for all of the styling is onerous for the user importing these components. The defaults for the blockquote were also overriding styling set by the user. As a user importing these components, I would expect code blocks and block quotes to inherit whatever CSS my app uses for code and blockquotes.

### Changes
- refactor(SlateAsInputEditor): remove styling props so user can style with own css

### Flags
- This requires the user to write their own CSS for html elements of `blockquote` and `pre`. However, I think this is more intuitive for the developer, and if they already have css for these, it will not be overridden. They also will not be required to pass in a bunch of props.


DianaLease:
> We'll need to make an issue for creating `css` in TSv2

Yes, I will create an issue and will add the css in my PR in TSv2 where I am updating the cicero-ui dependency.

PR #136, opened 2019-10-18
fix(SlateAsInputEditor): set data-key for headings - I135

archhhh:
Signed-off-by: Artyom <artychen13@unist.ac.kr>

# Issue #135  
Fixes issue described in  [issue #135](https://github.com/accordproject/markdown-editor/issues/135).
This bug caused the following issue in [@accordproject/template-studio-v2 issue #113](https://github.com/accordproject/template-studio-v2/issues/113). Passed data-key from Heading Node was not set in heading html tags, as a result causing bugs when scrolling.

### Related Issues
- [issue #135](https://github.com/accordproject/markdown-editor/issues/135)
- [@accordproject/template-studio-v2 issue #113](https://github.com/accordproject/template-studio-v2/issues/113)

PR #159, opened 2019-10-25
feat(*): inline hyperlink tooltip styling - I143

husseyexplores:
# Issue #143 
Positions the hyperlink popup inline, relative to the selected text.

### Changes
- Gets the selected text node via ~~DOM `getSelection()`~~ `slate-react` api then calculates the position of the node in the DOM using hyperlink popup and selected nodes' `x, y, width height`.

**Current behavior**:
The user selects text, clicks the hyperlink button. The popup opens right under the selection. The hyperlink form text field contains the selected text by default. And the url field gets focused. Once the form is closed, the focus returns back to the editor where it left off via `slate-react` api.

Here is a short video of the current behavior. I add a couple of hyperlinks. https://www.loom.com/share/b68349b630cc49409d0e0dad54c26a28

**Other Changes:** Added `lockText` and `DIVIDER` proptypes because the eslint was complaining. And Using `as` prop for `Heading` styled component instead of `createElement`. `as` prop is [supported by styled-component](https://www.styled-components.com/docs/api#as-polymorphic-prop) to render dynamic elements.

### Flags
Once the popup ones, the text gets de-selected. Because it is not possible to select the text and focus any input field at the same time. The browsers don't allow that AFAIK. You can observe this behavior in the above video. As soon as I click the hyperlink button, the text gets de-selected. (I'm referring only to the visuals here. Adding the hyperlink works fine as slate knows the last selected text)

The Google docs retains the selection though. Upon further investigating how were they doing that, it turns out that they use an empty div with absolute position right behind the selected text to mimic the native selection while the focus is on another input element. 

It seems not too difficult to implement at first. But if the selected text goes on two lines, then there are complications. 

GDocs wraps each line into a separate div, so if the selection goes on two or more lines, the simulated selection div can be positioned correctly on each line.

However, this is not the case in slate js. Each chunk of text stays in the same wrapper element even if it goes on multiple lines (Unless the block type changes) so we can't get the proper position of the selected text.

Does anyone know if it is possible to split each line into a separate div in slate? If it is possible, I can try to give this a shot as well, in a separate PR possibly?

### Related Issues
- Issue #100 
- Pull Request #142


irmerk:
Tagging @michael-grover

src/SlateAsInputEditor/index.js
```diff
@@ -63,20 +62,45 @@ const ToolbarWrapper = styled.div`
...
-);
+const Heading = styled.div`
+  font-family: serif;
+`
 
 Heading.propTypes = {
   type: PropTypes.oneOf(['h1', 'h2', 'h3', 'h4', 'h5', 'h6']),
 };
 
+const Code = styled.code(({ customCssObject }) => ({ ...customCssObject }))
```
DianaLease: We've decided to remove styling props for code, pre, and blockquote, so please remove all code related to this to match the current state of the master branch.

Michael-Grover:
This looks great! Make sure that when the user clicks the link button, the focus is changed to the "link URL" field such that their next keystroke will start typing in that field. I can't tell from your gif if it already does that. Check out how it works in Google docs. Let me know if you need more detail.

src/SlateAsInputEditor/index.js
husseyexplores: Sorry, I was having a problem signing the commits previously so I had to reopen a new PR and I mistakenly pulled the stale code also.
Just fixed it now.

husseyexplores:
@Michael-Grover Thanks for catching it. I updated the PR comments explaining the new behavior and also added a short video.

husseyexplores:
The scroll on focus issue is recently opened. (Issue #162). There is an open PR for that issue also (PR #163).

Should I remove the fix from this PR?

irmerk:
@DianaLease I think we need to review this in conjunction with #163

qualitymanifest:
> The scroll on focus issue is recently opened. (Issue #162). There is an open PR for that issue also (PR #163).
> 
> I'm removing the fix from here.

Oh yeah, I see you ended up at basically the same conclusion. The only downside I could find with doing nothing in the onFocus handler is that it causes you to lose focus when you tab into the editor. Other than that it solves the main issue.

husseyexplores:
Yes. You did much deeper tinkering to figure it out. 
I just wanted to see what exactly is happening while the editor gets focused. By adding my own onFocus handler with console logs, the problem went away so I added a noop function for the time being which I just removed after seeing your PR.

I also thought that it should be a separate issue and a PR for this for better issues and PR tracking.

qualitymanifest:
Nice, you found the solution very efficiently. I ended up chasing it across four repos haha

src/SlateAsInputEditor/index.js
```diff
@@ -63,15 +62,9 @@ const ToolbarWrapper = styled.div`
...
-      font-family: 'serif';
-    `,
-  {
-    'data-key': attributes['data-key']
-  },
-  children,
-);
+const Heading = styled.div`
+  font-family: serif;
+`
```
irmerk: I believe we needed this for the `data-key` in order to navigate through `cicero-ui`'s `Navigation` component

husseyexplores: I'm spreading the attributes like this
```
return <Heading as="h1" {...attributes}>{children}</Heading>;
```
Shouldn't this be enough as `attributes` object already contains the `data-key`?

husseyexplores: I can revert back to the `createElement` here also. I implemented this there was a focusing issue. But now the focus management is now explicit, via DOM and slate apis so I hope it will work with the previous version too.
I left it as is because it is shorter and simpler.

irmerk: I tried this locally and it still works... This comes from #92 and #136, I must just have gotten lost on why these changes needed to be made in the first place.

husseyexplores: Yes. #92 does pass any prop into createElement at all. And #136 fixes this by adding it explicitly.

So should I leave this as it is, right?

irmerk:
This all looks good to me so far. I want to review together with @DianaLease tomorrow.

src/SlateAsInputEditor/index.js
irmerk: I believe so, yes.

husseyexplores:
Since `Popop` and `Form` were the semantic ui components, they don't expose the DOM node. 
Semantic UI React provides a `Ref` component that gives us the DOM node reference via callback with `innerRef` prop so I used that.
I found the `Ref` example here: https://bit.dev/semantic-org/semantic-ui-react/ref

src/FormattingToolbar/index.js
```diff
@@ -190,6 +251,13 @@ export default class FormatToolbar extends React.Component {
     this.setState({ openSetLink: !this.state.openSetLink });
   }
 
+  /**
+   * Close set link form.
+   */
+  closeSetLinkForm() {
```
DianaLease: Sorry, just noticed this needs to binded in the constructor still.

src/FormattingToolbar/index.js
```diff
@@ -280,37 +348,109 @@ export default class FormatToolbar extends React.Component {
     );
   }
 
+  /**
+   * Calculates the link popup position and styles, if any
+   */
+  calculateLinkPopupPosition() {
```
DianaLease: This also should be binded in the contsructor.

src/FormattingToolbar/index.js
```diff
@@ -280,37 +348,109 @@ export default class FormatToolbar extends React.Component {
...
+      // and use our computed `top` and `left` values
+      popupStyle: { top, left, transform: 'none' },
+      popupPosition,
+    };
+  }
+
   /**
    * Render form in popup to set the link.
    */
   renderLinkSetForm() {
```
DianaLease: this also needs to be binded in the constructor

Issue #162, opened 2019-10-26
Unintentional scroll when editor receives focus


qualitymanifest:
See related https://github.com/accordproject/template-studio-v2/issues/78 for my past notes on this issue. Cicero-ui also has this issue and all of them can be fixed by a PR I will be sending over shortly

**To reproduce:**
- Open an editor and enter enough lines to cause some scrollbar
- Scroll down
- Click outside the editor, then click back into it, near the bottom of the editor

**The results:**
- The scroll position will jump.
- On Firefox, the cursor ends up at the beginning of the first line, even though that isn't where you clicked
- On Chrome, the cursor ends up where you clicked, but the window may have been scrolled to a point where the cursor is not visible
- Either way, after the initial buggy focus click, any subsequent clicks on the editor work as they should, until you focus outside the editor and come back.

**Expected behavior**
Cursor ends up where you click, no unnecessary scrolling happens

PR #163, opened 2019-10-26
fix(SlateAsInputEditor): fix focus issue by overriding Slate onFocus …

qualitymanifest:
Closes https://github.com/accordproject/markdown-editor/issues/162

Also closes https://github.com/accordproject/template-studio-v2/issues/78 once the new version of markdown-editor trickles up

See the above two issues for a description of the symptoms

I believe this issue is caused in the [slate `onFocus` method](https://github.com/ianstormtaylor/slate/blob/6269fcaa7168d6a7b79a62b286fd0d0316819c8f/packages/slate-react/src/plugins/dom/after.js#L384). When `deselect` is called, something strange happens where the selection is briefly incorrect and is considered to be at the beginning of the first line. Then, [`scrollToSelection`](https://github.com/ianstormtaylor/slate/blob/3dce916074e0294a5045f9c47f6932ac5841e947/packages/slate-react/src/utils/scroll-to-selection.js) is called and brings you up there, to the wrong place.

Giving the editor a custom `onFocus` method allows you to bypass the slate `onFocus` method, so `deselect` is never called.

A little more on `deselect`... The comments from the slate onFocus method mention that the deselect is to prevent the *old* selection from being set by [`updateSelection`](https://github.com/ianstormtaylor/slate/blob/d28f78c0607ad147f85ff4a363fa65d7ddfa7c8c/packages/slate-react/src/components/content.js#L211). If you call `editor.focus()` in the custom `onFocus` method, you can reproduce this in Firefox (but not Chrome), where if you click into the editor, then click out, then click back into the editor, the cursor goes back to the *last* place you clicked instead of the current place you clicked. 

By never calling `editor.focus()` you avoid this issue however you run into some issues tabbing into the editor. **Putting `editor.focus` inside a `setTimeout` seems to fix some timing issue and avoids the issues that `deselect` both prevents and causes. Focusing into and out of the editor by mouse or tab works fine. Selecting text inside the editor by mouse or keyboard works fine.** This is on both Chrome and Firefox on Linux.

The `next` parameter inside the custom `onFocus` method refers to slate's `onFocus` method. So, by never calling `next()` you avoid calling `deselect`. Also, the `next` parameter inside slate's `onFocus` method appears to be `undefined`, so I don't think that you're breaking an important chain of events by not calling `next()`.

Signed-off-by: cdr <qualitymanifest@gmail.com>


irmerk:
@DianaLease I think we need to review this in conjunction with #159

irmerk:
This all looks good to me so far. I want to review together with @DianaLease tomorrow.

qualitymanifest:
Unfortunately I haven't been able to nail down why it happens here but not in the example editors provided by Slate. The explanation above is as far as my understanding of the issue goes.

This repo is definitely not alone in this issue though. There are several issues [like this](https://github.com/ianstormtaylor/slate/issues/1551) [and this](https://github.com/ianstormtaylor/slate/issues/2911) from other Slate users having similar focus jump issues.

Since there are similar open issues and I couldn't say for sure where the problem originates, I figured I would hold off on opening any issues on Slate for now. Hopefully somebody can locate the source of the issue, I think this would work for the time being though.

qualitymanifest:
Just a heads up, I noticed that one of the related issues I mentioned was closed by [this PR](https://github.com/ianstormtaylor/slate/pull/3093), so when slate gets updated you can probably delete the focus handler I added. It sounds like there were a lot of breaking changes between 0.47.0 and 0.50.0 so that's probably going to be a pretty big task but I figured it'd be worth knowing.

Issue #132, opened 2019-10-17
Copy + Paste Works in lockText Mode


irmerk:
**Describe the bug**
Copy and pasting inside a clause template is currently being allowed during `lockText`

**To Reproduce**
1. Run locally within `cicero-ui`
  a. Download both `markdown-editor` and `cicero-ui`
  b. Run in `markdown-editor`
    - `npm install`
    - `npm run build`
    - `npm run transpile`
    - `npm link`
  c. Run in `cicero-ui`:
    - `npm install`
    - `rm -rf node_modules/\@accordproject/markdown-editor`
    - `npm link @accordproject/markdown-editor`
    - `npm run start`
2. Go to `http://localhost:8080/`
3. Ensure the `lockText` prop in `cicero-ui` being passed down to the `ContractEditor` is true.
4. Highlight text outside a clause (not a variable), and try pasting inside the clause.

**Expected behavior**
`lockText` should prevent any edits to text within a clause - except for variables

**Additional context**
Related to #122

michaelclark2:
I'd like to take a crack at this

irmerk:
@michaelclark2 any progress on this? Opening back up in the meantime.

michaelclark2:
I'm close, I need to possibly make pull requests to both repos

michaelclark2:
I've narrowed down the issue to the way you are adding the props.lockText as an annotation.  It gets unset somehow before it gets read in the isEditorLockText and I'm having trouble figuring out how to control that flow.  But figuring this out should prevent any changes to a clause while lockText === true

michaelclark2:
I've got it to where it will prevent any changes, but only after adding a single character.  I added a preventDefault in the onKeyDown, and the character doesn't get added when I'm using a breakpoint, so maybe there's some asynchronicity going on somewhere?

irmerk:
@michaelclark2 I think I know what you're referring to, #123. Is that what you're seeing? If so, yeah unsure what is causing that behavior at the moment.

I keep thinking there is a delay in setting the annotation and thus the first character is allowed, but skeptical of this theory.

michaelclark2:
@irmerk yep, that's it.  I'll try to focus more on the copy paste side of this issue, but #123 should fix all the lockText issues (maybe)

irmerk:
@michaelclark2 any progress on the copy/paste side of things?

DianaLease:
[...]

Issue #123, opened 2019-10-16
First Keypress in lockText Mode Executes


irmerk:
**Describe the bug**
Text formatting is currently being allowed during `lockText`

**To Reproduce**
1. Run locally within `cicero-ui`
  a. Download both `markdown-editor` and `cicero-ui`
  b. Run in `markdown-editor`
    - `npm install`
    - `npm run build`
    - `npm run transpile`
    - `npm link`
  c. Run in `cicero-ui`:
    - `npm install`
    - `rm -rf node_modules/\@accordproject/markdown-editor`
    - `npm link @accordproject/markdown-editor`
    - `npm run start`
2. Go to `http://localhost:8080/`
3. Ensure the `lockText` prop in `cicero-ui` being passed down to the `ContractEditor` is true.
4. Place selection within a clause, start typing. Notice the FIRST character key press executes.

**Expected behavior**
`lockText` should prevent any edits to text within a clause - except for variables

**Additional context**
N/A

rishabhx64:
Can I take this up?

pramodrhegde:
@irmerk Please assign this to me.

irmerk:
@pramodrhegde any progress on this?

pramodrhegde:
@irmerk Yes, I have started working on it. will raise a PR once I am ready with the fix.

pramodrhegde:
@irmerk Please review.

irmerk:
@pramodrhegde checking in on if there is any progress

PR #171, opened 2019-10-30
fix: First Keypress in lockText Mode Executes bug fix

pramodrhegde:
# Issue #123 
This PR fixes the bug wherein the first keypress in lockText mode was executing.

### Changes
- added `renderAnnotation` callback hook
  - this renders any annotations which are added/updated via useEffect hook (here for lockText annotation)

### Flags
- none

### Related Issues
- Issue #132

### DIFF
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -442,6 +442,13 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
     onChange(value);
   };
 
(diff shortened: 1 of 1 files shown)


irmerk:
Running this locally in `cicero-ui`, and I'm seeing the first character inside a clause is still allowed.

irmerk:
@pramodrhegde any progress on this?

irmerk:
Fixed in #202

Issue #156, opened 2019-10-25
Creating lists via toolbar throws error


DianaLease:
**Describe the bug**
Creating lists via toolbar throws error

**To Reproduce**
1. Highlight text
2. Click on ordered or unordered list icon
3. Open JS console

**Expected behavior**
Should create list and not throw error.

**Screenshots**
![Screen Shot 2019-10-25 at 1 54 00 PM](https://user-images.githubusercontent.com/20543103/67592924-ef7cca80-f72e-11e9-8e93-fa71b72b531e.png)

DianaLease:
When running locally, looks like the error is coming from concerto-core, which seems odd... @jeromesimeon 
![Screen Shot 2019-10-25 at 5 12 15 PM](https://user-images.githubusercontent.com/20543103/67604826-ac305500-f74a-11e9-9748-1fe3c00d46cf.png)

DianaLease:
Looks like it it happening when it runs the validation on lists [here](https://github.com/accordproject/concerto/blob/master/packages/concerto-core/lib/serializer/resourcevalidator.js). I'm confused as to what this validation should be applied to though since when we get this error, we are not in a concerto model. Let's sync up on this next week @jeromesimeon as I'm sure you understand this much better than I do.

PR #174, opened 2019-10-30
fix(schema): add tight property for lists

DianaLease:

# Issue https://github.com/accordproject/markdown-editor/issues/156

### Changes
fix(schema): add tight property for lists

### DIFF
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -241,7 +241,7 @@ export default class FormatToolbar extends React.Component {
       // Selection is not a list.
       } else {
         editor.withoutNormalizing(() => {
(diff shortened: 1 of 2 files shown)


jeromesimeon:
Looks good! FYI: The issue I opened in markdown-transform is actually about nested list in Slate (apparently they are gobbled). Do we support nested lists on the UI side? https://github.com/accordproject/markdown-transform/issues/109

DianaLease:
> Looks good! FYI: The issue I opened in markdown-transform is actually about nested list in Slate (apparently they are gobbled). Do we support nested lists on the UI side? [accordproject/markdown-transform#109](https://github.com/accordproject/markdown-transform/issues/109)

not currently, no.

PR #178, opened 2019-11-05
feat(schema): force next block after html to paragraph

irmerk:
# No Issue
Changes based on info from these two ([one][slack1] and [two][slack2]) conversations in SlateJS Slack

### Changes
- Force a block after `'html_block'` to be a `'paragraph'`
  - This should apply to `clause` blocks

### Flags
- Unsure how to make the previous block a paragraph, `previous` did not work in addition to `next`
- There may be things other than `clause` blocks using `'html_block'`

### Related Issues
- [TSv2 Issue #145][tsv2i]
- [TSv2 Pull Request #146][tsv2p]

[tsv2i]: https://github.com/accordproject/template-studio-v2/issues/145
[tsv2p]: https://github.com/accordproject/template-studio-v2/pull/146
[slack1]: https://slate-js.slack.com/archives/C1RH7AXSS/p1554819145354500
[slack2]: https://slate-js.slack.com/archives/C1RH7AXSS/p1555064971022200

### DIFF
--- a/src/schema.js
+++ b/src/schema.js
@@ -34,6 +34,8 @@ const schema = {
       ],
     },
     html_block: {
+      next: { type: 'paragraph' },
+      previous: { type: 'paragraph' },
       nodes: [
         { match: { type: 'paragraph' } },
       ]
--- a/package-lock.json
+++ b/package-lock.json
@@ -84,9 +84,9 @@
       }
     },
     "@babel/cli": {
-      "version": "7.6.4",
-      "resolved": "https://registry.npmjs.org/@babel/cli/-/cli-7.6.4.tgz",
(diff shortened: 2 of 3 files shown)


irmerk:
I believe `last` will make a paragraph at the end of the document, whereas `next` is only the following block. Not certain about this though.

dselman:
> I believe `last` will make a paragraph at the end of the document, whereas `next` is only the following block. Not certain about this though.

FYI `first` and `last` apply to the child nodes of the match. E.g. it allows you to say that the first child of an ol is a li. 
https://github.com/ianstormtaylor/slate/blob/8dd919dc3457c62ea2aa72f31b83faf42cbaba72/docs/reference/slate/schema.md

src/schema.js
```diff
@@ -34,6 +34,7 @@ const schema = {
       ],
     },
     html_block: {
+      next: { type: 'paragraph' },
```
dselman: You can use `previous` to indicate that the sibling node before the `html_block` is a paragraph.

irmerk:
Weird, `previous` didn't work for me yesterday and now does. I must have had an issue running locally yesterday.

Issue #187, opened 2019-11-08
Soft Break Error after Hyphen


irmerk:
**Describe the bug**
Parse error which may result from soft breaks after a hyphen?

**To Reproduce**
1. Go to the [TSv2 Netlify](https://accordproject-studio.netlify.com/)
2. Add `Volumediscountulist` to the contract
3. Click inside the clause

**Screenshots**
<img width="1676" alt="Screen Shot 2019-11-08 at 15 00 12 PM" src="https://user-images.githubusercontent.com/36460856/68508055-5f18ac80-023b-11ea-9d44-d2bb69aa4348.png">

**Additional context**
- [Cicero-UI PR #218](https://github.com/accordproject/cicero-ui/pull/218)
- [Template Specification](https://templates.accordproject.org/volumediscountulist@0.2.0.html)
  - Note that "`per-Transaction`" seems to have a soft break
- [Commit for soft break support](https://github.com/accordproject/markdown-transform/commit/b5a583dd11cd8ef28841c9945f6604ea5252efa3)

irmerk:
The [updates to `markdown-transform`][transUpdate] to handle the `softbreak` in a list are not working in [the PR for the schema][schemaPR], only [adding the `softbreak` to nodes match of `list_item`s][PRL] works still. Without this, paragraphs _should_ be enclosing `softbreaks`/`inlines`/`linebreaks` in `list_item`s, but I think something additional may be needed in the `markdown-transform` to account for this?

<img width="875" alt="Screen Shot 2019-11-12 at 10 25 47 AM" src="https://user-images.githubusercontent.com/36460856/68684727-f8dfa280-0536-11ea-8e92-8331f291b765.png">

[transUpdate]: https://github.com/accordproject/markdown-transform/releases/tag/v0.8.0
[schemaPR]: https://github.com/accordproject/markdown-editor/pull/188
[PRL]: https://github.com/accordproject/markdown-editor/pull/188/commits/60736db1b51444ed567be991e212698862196687#diff-1b21ae31202733dc0c9c48a31aa51040R95

jeromesimeon:
It's a little hard to make the connection, but AFAICT the Slate DOM out of the `markdown-slate` package looks fine, with a proper enclosing paragraph:
```
bash-3.2$ markus --version
0.8.0
bash-3.2$ markus parse --sample ~/git/cicero-template-library/src/volumediscountulist/text/sample.md --slate 
10:44:40 - info:
{
  "object": "value",
[...]

PR #188, opened 2019-11-11
fix(schema): support softbreaks in lists - I187

irmerk:
# Issue #187
Support for `softbreak`s in a `list_item`

### Changes
- Allow for `list_item`s to be a paragraph, not just `text` or `link`s
- Allow `softbreak`s in a list

### Flags
- N/A

### Related Issues
- N/A


src/schema.js
```diff
@@ -86,8 +86,21 @@ const schema = {
...
       parent: [{ type: 'ol_list' }, { type: 'ul_list' }],
-      nodes: [{ match: [{ object: 'text' }, { type: 'link' }] }],
-      marks: [{ type: 'bold' }, { type: 'italic' }, { type: 'code' }],
+      nodes: [
+        {
+          match: [
+            { object: 'text' },
+            { type: 'link' },
+            { type: 'paragraph' },
+            { type: 'softbreak' }
```
dselman: I don't think it is valid to just have a soft break as a list item. A soft break should be inside a paragraph.

irmerk: Unsure how to implement this. Currently, it is parsing correctly. If I remove `'softbreak'` from `list_item` and rely on `list_item` being a `'paragraph'`, which can be a `'softbreak'`, it does not parse correctly.

irmerk:
The current state of this code run in TSv2 still results in a parse error.

Issue #192, opened 2019-11-13
Slate Losing Lists


irmerk:
**Describe the bug**
There seems to be an error with the schema for SlateJS and lists are lost when coming from markdown to slate in the `markdown-transform`

**To Reproduce**
Roundtrip Slate DOM to markdown back to Slate DOM

**Expected behavior**
There should be rendering of list items correctly, not as paragraphs?

**Screenshots**
<img width="615" alt="Screen Shot 2019-11-13 at 14 53 30 PM" src="https://user-images.githubusercontent.com/36460856/68799036-7bdf2680-0625-11ea-948e-2a6a8a997326.png">
Note that the indentation is way off after a list now.

<img width="608" alt="Screen Shot 2019-11-13 at 14 53 42 PM" src="https://user-images.githubusercontent.com/36460856/68799040-7da8ea00-0625-11ea-8f32-ded084b8a6c4.png">
Note that this is not being registered as a list in the toolbar, and entering creates a plain paragraph (indented)

irmerk:
In the demo you can see that the list already created and there in the demo renders with a paragraph tag, but a new list created by the user has list items which are _not_ paragraph tags:

---

**Default list**:
<img width="1079" alt="Screen Shot 2019-11-14 at 17 23 38 PM" src="https://user-images.githubusercontent.com/36460856/68901355-c7680200-0703-11ea-9356-7c4fbe4d372f.png">

---

**User made list**:
<img width="1087" alt="Screen Shot 2019-11-14 at 17 23 13 PM" src="https://user-images.githubusercontent.com/36460856/68901338-c040f400-0703-11ea-9623-292c3c494990.png">

---

The default ones are not registering as a list to the toolbar because it is being recognized as just a paragraph.

jeromesimeon:
This looks like a regression introduced when we switched to markdown-transform 0.8.0.

irmerk:
Right, this seems to have to do with needing to update the way we handle lists in this component now. Because it used to be 
```
ol_list => list_item => text
```
But now is
```
ol_list => list_item => paragraph => text
```

jeromesimeon:
makes sense! The relationship between the schema and how we modify the DOM is also pretty sensitive. I'm wondering if there is a good way to document both so they remain in sync.

PR #194, opened 2019-11-15
fix(src): handle new form of list item as paragraph - I192

irmerk:
# Issue #192
Adjust the handling of lists with the new `paragraph` support in `markdown-transform v0.8`

### Changes
- Keep focus when clicking a block in the toolbar
- Handle block types as constants
- Handle breakout of empty list item
- Modularize and refactor code
  - Block action methods
  - `onClickBlock` handler
  - Handle lists `onEnter` in the plugin

### Flags
- Work still needed for this PR
  - ~Persisting issue around `block_quote` to `list_item`, requires further work~
  - ~Formatting Toolbar buttons don't yet recognize the selection being in a list~
- Future Issues
  - Work around `SHIFT + ENTER` handling a `linebreak` / new `paragraph` in a `list_item` is needed

### Related Issues
- N/A


irmerk:
### [Netlify deployment](https://deploy-preview-194--accordproject-markdown-editor.netlify.com/examples/)

src/plugins/list.js
```diff
@@ -11,7 +13,35 @@ function List() {
...
+      return false;
+    }
+
+    // Hitting enter on a non-empty list item will add a new list_item
+    if (isSelectionList(value) && startBlock.text.length !== 0) {
+      editor.withoutNormalizing(() => {
+        event.preventDefault();
+        editor.splitBlock().unwrapBlock(CONST.LIST_ITEM).wrapBlock(CONST.LIST_ITEM);
+      });
+      return true;
```
mttrbrts: What is the effect of this? Shouldn't it return `next()` instead?

irmerk: I think `next()` is resulting in there being a `paragraph` in the `list_item` and a new `paragraph` below, as if a sibling of the `list_item`.

src/schema.js
```diff
@@ -89,8 +89,6 @@ const schema = {
       nodes: [
         {
           match: [
-            { object: 'text' },
-            { type: 'link' },
```
jeromesimeon: Shouldn't this include lists as well?

src/plugins/list.js
```diff
@@ -11,7 +13,35 @@ function List() {
...
+    event.preventDefault();
+
+    // Hitting enter on a blank list item will break out of the enclosing list
+    if (isSelectionList(value) && startBlock.text.length === 0) {
+      editor.withoutNormalizing(() => {
+        event.preventDefault();
+        editor
+          .setBlocks(CONST.PARAGRAPH)
+          .unwrapBlock(CONST.LIST_ITEM)
+          .unwrapBlock(CONST.OL_LIST)
```
jeromesimeon: I do not understand the double unwrapping of first `OL_LIST` and then `UL_LIST`. Shouldn't this unwrap only one? (i.e., the first enclosing list?)

src/plugins/list.js
jeromesimeon: I'm confused about this as well. What should be the type returned by `onEnter` ?

src/schema.js
irmerk: I believe so, but I think that should be left for another issue when we add support for nested lists.

src/plugins/list.js
irmerk: I'm unsure if it should be returning a type. @dselman you may have some insight?

src/plugins/list.js
irmerk: Good point. I think Slate was handling this as one unwrap to whichever list it was in. I refactored to unwrap once for whichever list it was.

src/plugins/list.js
DianaLease: What is `.unwrapBlock(CONST.LIST_ITEM).wrapBlock(CONST.LIST_ITEM)` doing on line 37? Why would we unwrap a list item just to wrap it again?

src/plugins/list.js
```diff
@@ -11,7 +13,34 @@ function List() {
...
+          .unwrapBlock(currentList(value).type);
+      });
+      return false;
+    }
+
+    // Hitting enter on a non-empty list item will add a new list_item
+    if (isSelectionInput(value, CONST.LIST_ITEM) && startBlock.text.length !== 0) {
+      editor.withoutNormalizing(() => {
+        event.preventDefault();
+        editor.splitBlock().unwrapBlock(CONST.LIST_ITEM).wrapBlock(CONST.LIST_ITEM);
```
DianaLease: What is .unwrapBlock(CONST.LIST_ITEM).wrapBlock(CONST.LIST_ITEM) doing on line 37? Why would we unwrap a list item just to wrap it again?

irmerk: I made this diagram that hopefully helps explain the process.

Image from iOS (1)

DianaLease: Thanks for explaining this. Makes sense now!

Issue #199, opened 2019-11-06
Cannot edit a link


dselman:
**Describe the bug**
Once inserted a link cannot be edited, as pressing the link toolbar button removes the link.

**To Reproduce**
1. Press the link button to insert a link
2. Press the link button again, to edit the link
3. Note that the link has been removed

**Expected behavior**
Pressing the link button when a link has already been inserted should allow the user to edit the link. It would also be nice to be able to open the link (e.g. as Google Docs does).

PR #201, opened 2019-11-21
feat(FormattingToolbar): support editing current links - I199

irmerk:
# Issue #199
Support editing a current link with the popup

### Changes
- Use pointer on hover over a link
- Add logic for if you click on a link
  - Move selection to this node to highlight all of it
  - Calculate the middle of this selection so that the caret of the popup shows in the middle
  - Link the text the user selected with the text inside the popup
- Handle editing a current link by unwrapping the link, deleting, and starting from scratch with a new link

### Flags
- Using the keyboard, moving to select the link highlights it all and disables further keyboard navigation
- ~Editing a current link will lose the cursor place~
- ~Highlighting a whole section / paragraph that includes a link only selects the link~
- ~Making bold *and* italic text a link crashes~

### [Deployed Netlify](https://deploy-preview-201--accordproject-markdown-editor.netlify.com/examples/)


DianaLease:
This looks great! One thing I'm noticing when playing with it in the netlify is that _sometimes_ when I try to make/edit a second link, the text in the text box shows the text of the first link. I'm trying to pinpoint when exactly it happens...

irmerk:
New flag: Click an existing link to edit, do not edit and only press enter, it crashes.

dselman:
FYI on iOS (iPhone) clicking on a link moves the selection to the beginning or end of the link. Why do we need to modify the selection? Won’t this break keyboard navigation? For example, using right-arrow to move through the doc.

dselman:
Modelling this on how Google Docs handles links would be best IMO - they don't modify the selection on keyboard presses, but they display a hover below the link so that it can be both opened and edited.

Screenshot 2019-11-22 at 07 55 52

irmerk:
Another flag: There's a small edge case that when you are at the end of a link and `SHIFT + <-` to highlight the link, the popup does not have the url in it.

PR #202, opened 2019-11-22
fix(SlateAsInputEditor): fix bug where first key press edited non-editable text

DianaLease:

# Issue #123 
- fix(SlateAsInputEditor): fix bug where first key press edited non-editable text
- Because we now pass props all the way down to the editor, we can get the lockText prop off the slate editor itself. This works to be able to toggle lock text at runtime without the need to add annotations.


irmerk:
Having trouble running this locally to see, but this should allow the cicero-ui demo to switch `lockText` from true to false and back and the editor reading it?

DianaLease:
> Having trouble running this locally to see, but this should allow the cicero-ui demo to switch `lockText` from true to false and back and the editor reading it?

Yes!

PR #222, opened 2020-01-31
[WIP] feat(SlateAsInputEditor): adds id to headings to be used as anchors

DianaLease:

### Issue 
- As the editor stands today, there is no way to link to a specific place on the document

### Changes
- Adds an `id` property to headings which is a combination of the text and key
- This will allow us to link to a specific headings on the web page like so `http://localhost:3001/examples#Heading%20Two-35`

### Flags
- Because of the toolbar, the heading the page jumps to actually gets hidden
- I have not tested how this will work when the editor is imported into another app where it has a fixed height and its own scrollbar.
- Because the "key" property can change, it's possibly if the document is edited, the key of a heading could change and therefore the link would no longer point to that heading.
- Right now to create the link, you'd have to inspect the element and get the `id`. We should have an icon or something similar to what github does that the user can click and then get the link from the browser's url field. In the screenshot from github below, the link icon appears on hover. (cc @Michael-Grover)
Screen Shot 2020-01-31 at 10 22 59 AM

### DIFF
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -171,24 +171,25 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
    * Renders a block
    */
   // @ts-ignore
-  const renderBlock = useCallback((props, editor, next) => {
-    const { node, attributes, children } = props;
+  const renderBlock = useCallback((blockProps, editor, next) => {
+    const { node, attributes, children } = blockProps;
+    const anchorName = encodeURIComponent(`${node.text}-${node.key}`);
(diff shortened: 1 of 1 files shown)


elit-altum:
Hey @DianaLease, can I help in any way with this ? I would like to work to implement this.

irmerk:
@elit-altum I'd hold off on this for a few days, @dselman is migrating `markdown-editor` to Slate `v0.5x`, so this will likely be easier once that's done.

elit-altum:
Sure! Thank you for clarifying @irmerk !

PR #233, opened 2020-02-10
(fix) debugging backspace issues with hr

dselman:
Signed-off-by: Dan Selman <danscode@selman.org>

# Issue #<NUMBER HERE>

### Changes
- <INSERT DESCRIPTION OF CHANGES>
  - <SUB NOTES OF DESCRIPTIONS>
- <SUMMARIZE ALL COMMITS IN PULL REQUEST>

### Flags
- <DESCRIBE ISSUES OR HOLDS FOR REVIEWERS>

### Related Issues
- Issue #<NUMBER HERE>
- Pull Request #<NUMBER HERE>

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -13,6 +13,15 @@ body {
     text-indent: 0ex;
 }
 
+.hr {
+    overflow: visible; /* For IE */
+    padding: 0;
+    border: none;
+    border-top: medium double #333;
+    color: #333;
+    text-align: center;
+}
+
 a {
     cursor: pointer;
 }
--- a/examples/src/index.js
+++ b/examples/src/index.js
@@ -33,6 +33,12 @@ linebreak.
 ## Heading Two
 This is more text.
 
+Before a thematic break.
+
+---
+
+After a thematic break.
+
 Ordered lists:
 
 1. one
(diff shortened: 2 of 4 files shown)


jeromesimeon:
I was testing by deleting from the end of the document. The delete skipped over the thematic break.
I think it worked if I delete from the previous paragraph.

dselman:
> I was testing by deleting from the end of the document. The delete skipped over the thematic break.
> I think it worked if I delete from the previous paragraph.

This seems to work for me in Safari and Chrome.
https://deploy-preview-233--accordproject-markdown-editor.netlify.com/examples/

jeromesimeon:
Yes, I think it's only when the thematic break is the last node in the document? it's a bit of an edge case.

BackspaceThematicBreak

dselman:
> Yes, I think it's only when the thematic break is the last node in the document? it's a bit of an edge case.
> 
> BackspaceThematicBreak

Interesting. I can reproduce that on Netlify, but not locally.

dselman:
I can also reproduce this with images. It seems to be an issue deleting void (?) blocks.

jeromesimeon:
Could this be in the logic that attempts to add an empty text node at the end?

src/styles.css
```diff
@@ -13,6 +13,15 @@ body {
     text-indent: 0ex;
 }
 
+.hr {
+    overflow: visible; /* For IE */
+    padding: 0;
+    border: none;
+    border-top: medium double #333;
```
DianaLease: why double? do we want an `hr` to appear as two lines?

dselman: I've no opinion on that... Whatever people prefer. The bug seems to be fixed moving from an `<hr>` to a `<div>` --- not really sure *why* and we have the same issue with the `<image/>` tag.

Issue #225, opened 2020-02-06
Hitting enter in a block quote is not handled as expected


DianaLease:
**Describe the bug**
Hitting enter in a block quote is not handled as expected

**To Reproduce**
1. Add a block quote
2. Hit enter to create a new line inside of the block quote
3. Inspect slate dom
4. See that the new line was created inside of the same node

**Expected behavior**
- The slate dom created should be a softbreak node followed by a new text node inside of the same paragraph inside of the blockquote.
- With [this change](https://github.com/accordproject/markdown-transform/pull/158) to the markdown-transform, this should make sure the markdown conversion roundtrips.
~~- As per @Michael-Grover - Look at how others handle block quotes. In Medium, shift + enter keeps you inside the quote while enter breaks out. Slab moves to next line on enter and breaks out on double enter.~~ See https://github.com/accordproject/markdown-editor/issues/218.

**Notes**
- Create `blockQuote` plugin with method similar to that in the list plugin https://github.com/accordproject/markdown-editor/blob/master/src/plugins/list.js#L16

Issue #218, opened 2019-12-20
Escape Block Quote


irmerk:
**Describe the bug**
If a block quote is all that exists, the user cannot escape with the keyboard.

**To Reproduce**
1. Run `markdown-editor` locally (or within `cicero-ui`)
2. Remove all text
3. Toggle block quote
4. You are now unable to escape from the block quote (save for the toolbar functionality)

**Expected behavior**
Similar logic to exiting a list. Backspace in an empty block quote should remove it, enter twice should exit.

PR #234, opened 2020-02-10
feat(blockquote): add ability to break out of block quote - 218

DianaLease:

# Issue #218 

### Changes
- feat(blockquote): add ability to break out of block quote

I followed the same behavior that Slab has:
- Hitting enter empty line will break out of the block quote
- Backspacing on the last item when empty will break out of the block quote

### Related Issues
#225 - The slate dom portion of issue 225 issue is giving me more trouble than I expected, so I will leave that for a separate PR (cc @jeromesimeon)

### DIFF
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -31,6 +31,7 @@ import baseSchema from '../schema';
 import PluginManager from '../PluginManager';
 import FormatToolbar from '../FormattingToolbar';
 import ListPlugin from '../plugins/list';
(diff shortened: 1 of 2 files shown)


DianaLease:
@dselman where are you referring to? The editor will always use this plugin so I do not think third party users should ever need to import it themselves unless I'm missing something.

dselman:
> @dselman where are you referring to? The editor will always use this plugin so I do not think third party users should ever need to import it themselves unless I'm missing something.

You're right - not necessary for this one.

Issue #213, opened 2019-12-18
Ability to open links


jeromesimeon:
**Is your feature request related to a problem? Please describe.**
I think it would be nice to be able to open links from the editor.
As far as I can tell right now there is only a way to edit a link, not to open it.

<img width="333" alt="Screenshot 2019-12-18 at 5 04 55 PM" src="https://user-images.githubusercontent.com/670099/71127758-a76fa680-21b9-11ea-96e4-ec45986b3349.png">

**Describe the solution you'd like**
Ability to open a link (through a click on the link itself or a button).

jeromesimeon:
For instance in Google docs:

<img width="669" alt="Screenshot 2020-02-12 at 9 35 01 AM" src="https://user-images.githubusercontent.com/670099/74344711-034e4a00-4d7b-11ea-83f2-ccba6681fbc8.png">

In order from left to right you can:
1. Click to open the link in a new tab
2. Copy the link
3. Edit the link
4. Remove the link

98lenvi:
Can I try?

dselman:
> Can I try?

Assigned to you. Thanks!

dselman:
Google Docs is another source of inspiration.

PR #246, opened 2020-02-29
feat(FormattingToolbar): ability to open links - #213

98lenvi:
# Issue #213 
Modified popover when a link is clicked on

### Changes
- The popover will have two UI, one when there is a link present in the text selected and others when there is no link
  - When there is a link, it will show 3 buttons with the actions: Copy, Edit, Un-Link
  - When there is no link, the previous Forn will be shown, i.e. A text-field for `href` and text with two action buttons, i.e. Save and unlink
  - I have Added a state variable - `showLinkForm` to decide which UI should be shown, but the popover also depends on whether the selected is a link or not.
  - The state variable was also added because I needed to prevent the focusing of the elements in the form when the form wasn't being rendered.
- The changes show a popover similar to Google Docs

Demo -
Peek 2020-03-02 20-28

### Flags
- Link preview (metadata / og metadata) couldn't be implemented like Google Docs because of CORS.
- The way the link is being displayed needs modifications & feedback

### Related Issues
- Issue #213

### DIFF
--- a/src/FormattingToolbar/LinkComponent/index.js
+++ b/src/FormattingToolbar/LinkComponent/index.js
@@ -59,7 +59,7 @@ const calculateLinkPopupPosition = (editor, openSetLink, setLinkFormPopup) => {
   return {
     // Disable semantic ui popup placement by overriding `transform`
     // and use our computed `top` and `left` values
-    popupStyle: { top, left, transform: 'none', width:'400px' },
+    popupStyle: { top, left, transform: 'none', width:'400px', },
     popupPosition,
   };
 };
--- a/src/FormattingToolbar/LinkComponent/index.js
+++ b/src/FormattingToolbar/LinkComponent/index.js
@@ -20,7 +20,7 @@ const calculateLinkPopupPosition = (editor, openSetLink, setLinkFormPopup) => {
     return {
       popupPosition,
       // Hide the popup by setting negative zIndex
-      popupStyle: { zIndex: -1 }
+      popupStyle: { zIndex: -1}
     };
   }
 
@@ -59,7 +59,7 @@ const calculateLinkPopupPosition = (editor, openSetLink, setLinkFormPopup) => {
   return {
     // Disable semantic ui popup placement by overriding `transform`
     // and use our computed `top` and `left` values
-    popupStyle: { top, left, transform: 'none' },
+    popupStyle: { top, left, transform: 'none', width:'400px' },
     popupPosition,
   };
 };
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -3,7 +3,7 @@ import { createPortal } from 'react-dom';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import {
-  Button, Dropdown, Form, Input, Popup, Ref, Icon, Grid
+  Button, Dropdown, Form, Input, Popup, Ref
 } from 'semantic-ui-react';
 
 import * as action from './toolbarMethods';
@@ -91,7 +91,7 @@ const DropdownHeader3 = {
 export default class FormatToolbar extends React.Component {
   constructor() {
     super();
-    this.state = { openSetLink: false, showLinkForm: true };
+    this.state = { openSetLink: false };
     this.linkButtonRef = createRef();
     this.hyperlinkInputRef = createRef();
     this.onMouseDown = this.onMouseDown.bind(this);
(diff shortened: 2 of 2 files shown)


dselman:
This is looking very promising! I think the design needs a little work @Michael-Grover may be able to help with that.

irmerk:
@mttrbrts tagging for input on this:
> Link preview (metadata / og metadata) couldn't be implemented like Google Docs because of CORS.

irmerk:
Subject to feedback from @Michael-Grover, I think the two popups should have the same style?

98lenvi:
Okie then! I'll wait for some more inputs on this.

mttrbrts:
> Link preview (metadata / og metadata) couldn't be implemented like Google Docs because of CORS.

I think that not supporting this in the library is perfectly acceptable. Google seems to have local caches of `og:image` on `gstatic.com`, so that they can trust the domain. It should be up to client applications to determine their CORS policy and how to handle image storage.

Michael-Grover:
@98lenvi the Netlify preview isn't working, can you provide a video or screenshots?

jeromesimeon:
> @98lenvi the Netlify preview isn't working, can you provide a video or screenshots?

@Michael-Grover it's here: https://deploy-preview-246--accordproject-markdown-editor.netlify.com/examples/

98lenvi:
@jeromesimeon @Michael-Grover The Netlify preview isn't working for me as well. I'll attach the screen recording shortly.

irmerk:
This is a longstanding issue with this repository. Netlify URLs need to have "/examples" added to the end.

jeromesimeon:
> @jeromesimeon @Michael-Grover The Netlify preview isn't working for me as well. I'll attach the screen recording shortly.

I have a PR https://github.com/accordproject/markdown-editor/pull/253 for fixing the netlify preview (it's there but in the `./examples` directory, which is inconvenient for PR reviews).

98lenvi:
I have updated the pull request description with a demo gif @Michael-Grover . Please check it there!

Michael-Grover:
I think the sliding icons are too elaborate. When a user clicks on an existing link, it should look the same as creating a new link, except with a link the user can click on that opens the link in a new tab. @jeromesimeon what do you think?

image

The link could be truncated like Google
image

Can we also make the link editor box 400px wide to accommodate longer links?

jeromesimeon:
> I think the sliding icons are too elaborate. When a user clicks on an existing link, it should look the same as creating a new link, except with a link the user can click on that opens the link in a new tab. @jeromesimeon what do you think?
> 
> image
> 
> The link could be truncated like Google
> image
> 
> Can we also make the link editor box 400px wide to accommodate longer links?

I think I agree with @Michael-Grover . Would an "Open" button make sense? As in Google docs?

Michael-Grover:
@jeromesimeon I don't see that button when I look at a link in Google Docs

98lenvi:
@Michael-Grover @jeromesimeon If I make the changes as per the suggestions, where do I place the copy button(if necessary)

Michael-Grover:
@98lenvi let's leave out a copy button. For now I would like to stick to the original issue: adding the ability to open links in the link editor.

98lenvi:
@Michael-Grover. I have updated it, this is how it looks now. Please check it
Peek 2020-03-04 14-54

src/FormattingToolbar/LinkComponent/index.js
```diff
@@ -59,7 +59,7 @@ const calculateLinkPopupPosition = (editor, openSetLink, setLinkFormPopup) => {
   return {
     // Disable semantic ui popup placement by overriding `transform`
     // and use our computed `top` and `left` values
-    popupStyle: { top, left, transform: 'none' },
+    popupStyle: { top, left, transform: 'none', width:'400px' },
```
irmerk: Unsure we want this, could you explain why this was added?

src/FormattingToolbar/index.js
```diff
@@ -368,61 +368,89 @@ export default class FormatToolbar extends React.Component {
...
-
+    const calculateTruncatedLinkText = (text) =>{
+      let truncatedText;
+      try{
+        truncatedText = text.slice(0,20)+(text.length>20?'...':'');
+      }catch(_){
+        truncatedText = 'link';
+      }
+      return truncatedText;
+    }
```
irmerk: I believe this can be done with [CSS alone](https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow)

src/FormattingToolbar/index.js
```diff
@@ -368,61 +368,89 @@ export default class FormatToolbar extends React.Component {
...
+console.log(popupStyle)
```
irmerk: Want to remove this.

src/FormattingToolbar/LinkComponent/index.js
98lenvi: > I think the sliding icons are too elaborate. When a user clicks on an existing link, it should look the same as creating a new link, except with a link the user can click on that opens the link in a new tab. @jeromesimeon what do you think?
> 
> image
> 
> The link could be truncated like Google
> image
> 
> Can we also make the link editor box 400px wide to accommodate longer links?

@Michael-Grover wanted it to be `400px` wide so I added XD.

src/FormattingToolbar/index.js
98lenvi: Sure! Thank you so much for sharing the resource.

src/FormattingToolbar/index.js
98lenvi: I'm sorry about that. Will remove that

src/FormattingToolbar/LinkComponent/index.js
irmerk: OH didn't see that 👀

98lenvi:
@irmerk Please check the latest commit. Thank you :100:

Issue #248, opened 2020-02-29
Performance Boost for Key Inputs


irmerk:
We could likely refactor how the editor handles a key press to more efficiently determine how to perform. Right now we test the key press event against all possibilities before moving on.

### [The code in question](https://github.com/accordproject/markdown-editor/blob/master/src/SlateAsInputEditor/index.js#L377-L416):

```js
  /**
  * Called upon a keypress
  * @param {*} event
  * @param {*} editor
  * @param {*} next
  */
  const onKeyDown = async (event, editor, next) => {
    if (isHotKey('mod+z', event)) {
      if (editor.props.editorProps.onUndoOrRedo) {
        await editor.undo();
        return editor.props.editorProps.onUndoOrRedo(editor);
      }
    }
    if (isHotKey('mod+shift+z', event)) {
      if (editor.props.editorProps.onUndoOrRedo) {
        await editor.redo();
        return editor.props.editorProps.onUndoOrRedo(editor);
      }
    }
    if (isHotKey("mod+b", event)) {
      return editor.toggleMark("bold");
    }
    if (isHotKey("mod+i", event)) {
      return editor.toggleMark("italic");
    }
    if (isHotKey("mod+alt+c", event)) {
      return editor.toggleMark("code");
    }
    if (isHotKey("mod+q", event)) {
      return handleBlockQuotes(editor)
    }
    if (isHotKey("mod+shift+7", event)) {
      return handleList(editor, "ol_list");
    }
    if (isHotKey("mod+shift+8", event)) {
      return handleList(editor, "ul_list");
    }
    switch (event.key) {
      case 'Enter':
        return handleEnter(event, editor, next);
      case 'Backspace':
        return handleBackspace(event, editor, next);
      default:
        return next(); // allow
    }
  };
```

irmerk:
Reopening for further refinement: I was thinking of an object literal instead of a switch case - though I'm not 100% it's possible in this application.

A couple resources: 
- https://ultimatecourses.com/blog/deprecating-the-switch-statement-for-object-literals
- https://dev.to/lukyhenson/replace-your-switch-statement-and-multiple-if-and-else-using-object-literals-en-us-1dec

PR #254, opened 2020-03-02
enchancement(SlateAsInputEditor): change multiple if condition to sin…

98lenvi:
…gle switch - #248

Signed-off-by: lenvi <lenvin@oykuapp.com>

# Issue #248 
Changed multiple if conditions to a single switch

### Changes
- I went with a single switch instead of two nested switches for better readability

### Flags
- Let me know if there is a more efficient way. To my best knowledge, I came up with this

### Related Issues
- Issue #248

### DIFF
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -375,45 +375,45 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
   * @param {*} next
   */
   const onKeyDown = async (event, editor, next) => {
-    if (isHotKey('mod+z', event)) {
-      if (editor.props.editorProps.onUndoOrRedo) {
+    switch (true) {
+      
+      case event.key==='Enter':
+        return handleEnter(event, editor, next);  
+        
+      case event.key==='Backspace':
+        return handleBackspace(event, editor, next);
+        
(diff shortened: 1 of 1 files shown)


irmerk:
This is definitely an improvement. I was thinking of an object literal instead of a switch case - though I'm not 100% it's possible in this application.

A couple resources: 
- https://ultimatecourses.com/blog/deprecating-the-switch-statement-for-object-literals
- https://dev.to/lukyhenson/replace-your-switch-statement-and-multiple-if-and-else-using-object-literals-en-us-1dec

98lenvi:
Wow @irmerk , Thanks for the resource . I never knew about object literal . Thank you once again

irmerk:
Sure thing. Try that out and see if we can get rid of this switch case.

98lenvi:
@irmerk I tried but wasn't able to get rid of the switch. I'll try it later with a clear mind

elit-altum:
@98lenvi if its okay with you, can I give this a shot?

DianaLease:
This looks good to me! I'd say maybe we can merge this and then if someone wants to do further refactoring they can go from there?

98lenvi:
@elit-altum, I think you can open a new issue and try it out. Please do mention me in that issue, I'd love to know the solution/approach taken. Thanks!

irmerk:
I reopened #248 if you want to continue @elit-altum

Issue #258, opened 2020-03-03
Shortcut commands in Tooltips missing


shrey-sachdeva2000:
**Is your feature request related to a problem? Please describe.**
- With the addition of shortcut commands, they should also be mentioned in `toolbarTooltip` 
- Make the Tooltip more descriptive

**Describe the solution you'd like(reference)**
![image](https://user-images.githubusercontent.com/47667325/75786757-c7137580-5d8b-11ea-8645-61982fb6a286.png)

**Additional context**
- This feature continues #124
- The shortcut commands should be based on OS of user(Ctrl for Windows, ⌘ for Mac)

PR #247, opened 2020-02-29
style(FormattingToolbar, icons): improve markdown-editor UI/UX - I124

shrey-sachdeva2000:
Signed-off-by: Shrey Sachdeva <shreysachdeva.2000@gmail.com>

# Issue #124 
Polish markdown-editor overall UI/UX

### Proposed Changes
- Cleaning the toolbar with icons having equal proportions and similar style
- Increasing the editor width with more spacing for a cleaner look
- Removing header styles from dropdown for better accessibility
- Removing border color and radius that made the design look slightly unprofessional

### Flags
- Note that the changes will simultaneously need to be commited in `cicero-ui` and further in Clause template.

### Reference

UI

PR #259, opened 2020-03-03
feat(toolbarTip): added shortcut commands in tooltip based on OS - I258

shrey-sachdeva2000:
Signed-off-by: Shrey Sachdeva <shreysachdeva.2000@gmail.com>

# Issue #258 

### Changes
- Added shortcut commands in `toolbarTooltip` based on OS of user
- Made the quote and link formatting options more descriptive

### Screenshot
tooltip

### Flags
- Requires testing on MacOS and Linux platforms

### Related Issues
- Issue #247

### DIFF
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -675,7 +675,7 @@ export default class FormatToolbar extends React.Component {
             undoIcon.width(),
             undoIcon.padding(),
             undoIcon.vBox(),
-            'Undo (' + tips.MOD + '+Z)',
+            `Undo (${tips.MOD} +Z)`,
             'toolbar-2x2'
           )
       }
@@ -687,7 +687,7 @@ export default class FormatToolbar extends React.Component {
             redoIcon.width(),
             redoIcon.padding(),
             redoIcon.vBox(),
-            'Redo (' + tips.MOD + '+Y)',
+            `Redo (${tips.MOD} +Y)`,
             'toolbar-2x3'
           )
       }
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -451,7 +451,7 @@ export default class FormatToolbar extends React.Component {
 
     return (
       <Popup
-        content={'Hyperlink'}
+        content='Insert a link'
         style={style}
         position='bottom center'
         trigger={
@@ -489,7 +489,7 @@ export default class FormatToolbar extends React.Component {
 
     return (
       <Popup
-        content={tips.capitalizeWord(action)}
+        content={tips.capitalizeWord(action)} 
         style={style}
         position='bottom center'
         trigger={
@@ -648,7 +648,7 @@ export default class FormatToolbar extends React.Component {
             undoIcon.width(),
             undoIcon.padding(),
             undoIcon.vBox(),
-            'undo',
+            'Undo(' + tips.MOD + '+Z)',
             'toolbar-2x2'
           )
       }
@@ -660,7 +660,7 @@ export default class FormatToolbar extends React.Component {
             redoIcon.width(),
             redoIcon.padding(),
             redoIcon.vBox(),
-            'redo',
+            'Redo(' + tips.MOD + '+Y)',
             'toolbar-2x3'
           )
       }
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -675,7 +675,7 @@ export default class FormatToolbar extends React.Component {
             undoIcon.width(),
             undoIcon.padding(),
             undoIcon.vBox(),
-            `Undo (${tips.MOD} +Z)`,
+            'Undo (' + tips.MOD + '+Z)',
             'toolbar-2x2'
           )
       }
@@ -687,7 +687,7 @@ export default class FormatToolbar extends React.Component {
             redoIcon.width(),
             redoIcon.padding(),
             redoIcon.vBox(),
-            `Redo (${tips.MOD} +Y)`,
+            'Redo (' + tips.MOD + '+Y)',
             'toolbar-2x3'
           )
       }
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -675,7 +675,7 @@ export default class FormatToolbar extends React.Component {
             undoIcon.width(),
             undoIcon.padding(),
             undoIcon.vBox(),
-            'Undo(' + tips.MOD + '+Z)',
+            'Undo (' + tips.MOD + '+Z)',
             'toolbar-2x2'
           )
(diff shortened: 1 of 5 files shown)


98lenvi:
Working on Linux!

shrey-sachdeva2000:
> Working in Linux!

Great! Thanks for the review.

shrey-sachdeva2000:
> LGTM.
> Just a suggestion, maybe you could add a space between the action name and the parentheses enclosing the shortcut on the tool tip.

Yes, the spacing can be fixed here. Thanks for the review!

src/FormattingToolbar/toolbarTooltip.js
```diff
@@ -5,28 +5,59 @@
...
+    os = 'Windows';
+    MOD = 'Ctrl';
+  } else if (!os && /Linux/.test(platform)) {
+    os = 'Linux';
+    MOD = 'Ctrl';
+  }
+  return MOD; 
+}
+
+export const MOD = OS();
```
irmerk: Some feedback on this:

- It doesn't look like the variable `userAgent` is used.
- `macosPlatforms` and `windowsPlatforms` could potentially be objects with `os` and `MOD` values, so this function could be more functional and return one value rather than assigning a value and returning it.
- I think `let` and `const` should be preferred to `var`
- Could destructure `platform` (`const { platform } = window.navigator;`)

irmerk:
Looks great! Provided some feedback on the main function.

Michael-Grover:
Please change "Unordered list" and "Ordered list" back to "Bulleted List" and "Numbered List". That change is not part of the issue or PR description. 

Only developers know the terms "ordered list" and "unordered list". We should stick to the convention that is more widely understood and used in common editors like Word and Google Docs.

src/FormattingToolbar/toolbarTooltip.js
shrey-sachdeva2000:
```
function OS () {
  const {platform} = window.navigator;

  const macosPlatforms = {
    os: ["Macintosh", "MacIntel", "MacPPC", "Mac68K"],
    MOD: "⌘"
  };

  const windowsPlatforms = {
    os : ["Win32", "Win64", "Windows", "WinCE"],
    MOD: "Ctrl"
  };

  if (macosPlatforms.os.indexOf(platform) !== -1) {
    return macosPlatforms.MOD;
  }  else if (windowsPlatforms.os.indexOf(platform) !== -1) {
    return windowsPlatforms.MOD;
  } else if (!os && /Linux/.test(platform)) {
    return windowsPlatforms.MOD;
  }
}

export const MOD = OS();
```

shrey-sachdeva2000: Thanks for the review @irmerk. I hope this is what you meant. Please let me know if there's anything else that needs to be changed.

shrey-sachdeva2000:
> Please change "Unordered list" and "Ordered list" back to "Bulleted List" and "Numbered List". That change is not part of the issue or PR description.
> 
> Only developers know the terms "ordered list" and "unordered list". We should stick to the convention that is more widely understood and used in common editors like Word and Google Docs.

Sure, I'll make these changes.

src/FormattingToolbar/toolbarTooltip.js
shrey-sachdeva2000: Thanks for the review @irmerk. I'll make these changes and request for review again.

src/FormattingToolbar/toolbarTooltip.js
```diff
@@ -5,28 +5,60 @@
...
+  if (macosPlatforms.os.indexOf(platform) !== -1) {
+    return macosPlatforms.MOD;
+  }  else if (windowsPlatforms.os.indexOf(platform) !== -1) {
+    return windowsPlatforms.MOD;
+  } else if (!os && /Linux/.test(platform)) {
+    return windowsPlatforms.MOD;
+  }
+}
+
+export const MOD = OS();
```
irmerk: I'm thinking this can still go further, like:

```js
  const macosPlatforms = {
    os: {
      "Macintosh": true,
      "MacIntel": true,
      "MacPPC": true,
      "Mac68K": true
    },
    MOD: "⌘"
  };
```

You could then not need `.indexOf` which needs to iterate through the array, but rather look up by key value.

shrey-sachdeva2000:
```
function OS () {
  const {platform} = window.navigator;

  const macosPlatforms = {
    os: {
      "Macintosh": true, 
      "MacIntel": true, 
      "MacPPC": true, 
      "Mac68K": true
    },
    MOD: "⌘"
  };

  const windowsPlatforms = {
    os : {
      "Win32": true, 
      "Win64": true, 
      "Windows": true, 
      "WinCE": true
    },
    MOD: "Ctrl"
  };

  if (macosPlatforms.os[platform] !== undefined) {
    return macosPlatforms.MOD;
  }  else
    return windowsPlatforms.MOD;
}

```
@irmerk would this be a correct implementation of the same? I works fine though.

irmerk: Yes I believe so: 

```js
 if (macosPlatforms.os[platform]) return macosPlatforms.MOD;
 if (windowsPlatforms.os[platform]) return windowsPlatforms.MOD;
 return windowsPlatforms.MOD;
```

shrey-sachdeva2000: Yupp, thanks. I'll push the changes

shrey-sachdeva2000:
> This looks great to me. @DianaLease?
> 
> Although now I'm realizing if it's between ⌘ and Ctrl, the `os` key value doesn't need to be made in `windowsPlatforms`. Really, both `MOD` values don't need to be made, the function can return `'⌘'` or `'Ctrl'`

Oops! That's right. We really refactored this one to the core! I'll push another commit.

src/FormattingToolbar/index.js
```diff
@@ -675,7 +675,7 @@ export default class FormatToolbar extends React.Component {
             undoIcon.width(),
             undoIcon.padding(),
             undoIcon.vBox(),
-            'undo',
+            'Undo (' + tips.MOD + '+Z)',
```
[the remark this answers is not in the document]
shrey-sachdeva2000: Sure @DianaLease. I should've done that earlier itself. Thanks!

shrey-sachdeva2000:
The multiple commits are because I accidentally pushed a commit without the DCO signoff and when I tried to fix it through the suggested commands, it failed again because the unsigned commit wasn't the last one. The same issue can be found [here](https://github.com/probot/dco/issues/121). Apologies for the same.

shrey-sachdeva2000:
Please review @DianaLease. Thanks!

PR #260, opened 2020-03-06
refactor(Slate): hotkey switch-case to object literal

elit-altum:
# Issue #248 

Continues on the refactoring of Slate Inputs to object literals. 
The Hotkeys currently require a complex condition matching so they are still being handled by switch case while the rest of the conditions have been refactored to Object literals as they are safer and more comprehensive.

### Changes
- Refactored three conditions to Object literals

### Related Issues
- Issue #248

### DIFF
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -375,43 +375,58 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
   * @param {*} next
   */
   const onKeyDown = async (event, editor, next) => {
-    switch (true) {
-      
-      case event.key==='Enter':
-        return handleEnter(event, editor, next);  
-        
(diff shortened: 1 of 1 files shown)


elit-altum:
@irmerk Please review!

98lenvi:
Nice! , I had a question, wouldn't this increase the time complexity? Because first the object literal is checked, then again in the switch condition.

elit-altum:
> 
> 
> Nice! , I had a question, wouldn't this increase the time complexity? Because first the object literal is checked, then again in the switch condition.

I think they are not exactly 'checked'. In the object literal , according to the property, its function is directly replaced. Therefore the switch case complexity for hotkeys remains same but for the first two keys it is optimized.

98lenvi:
@elit-altum Thanks for clarifying sharing the knowledge!

PR #250, opened 2020-03-01
[WIP] feat(schema,ToolbarMethods): change in schema to allow nested list/quotes - #241

98lenvi:
Signed-off-by: lenvi <lenvin@oykuapp.com>

# Issue #241 
Initial steps towards enabling nested quotes/list

### Changes
- Change in `Schema` & `ToolbarMethods`
  - `transformBlockQuoteToList` no longer unwraps the `block_quote` attribute, it justs wraps a `ol_list` (or) `ul_list` and `list_item` attributes over the existing.
  - `transformListToBlockQuote` no longer unwraps the `list_item` and `ol_list` (or) `ul_list` attribute, it justs wraps a `block_quote` attributes over the existing.
  - The modifications to the above two methods were only possible because of the changes I made in the `schema`, without `schema` changes the text would just disappear.

### Flags
- I want to know if changes to the `schema` are acceptable? If the current approach sounds good then I'll proceed
- I have commented out the code which isn't required anymore.

### Related Issues
- Issue #241

Issue #241, opened 2020-02-17
Editing for nested quotes and lists


jeromesimeon:
**Is your feature request related to a problem? Please describe.**
Markdown can support nested quotes and lists. Those seem to be properly imported in the editor, but one cannot edit those or create them directly using the toolbar.

**Describe the solution you'd like**
It would be nice to be able to create a quote containing a list or a list containing a quote, or a quote containing a list containing a quote. This might require some thinking about the scope for the toolbar buttons.

jeromesimeon:
Currently one can load the following markdown:

<img width="1483" alt="Screenshot 2020-02-17 at 11 17 37 AM" src="https://user-images.githubusercontent.com/670099/74670433-290e9100-5177-11ea-980c-3bef8c720e8e.png">

Clicking on the 'list' button for the paragraph does two things: remove the quote and add the list. Should it just add the list without removing the quote?

<img width="1468" alt="Screenshot 2020-02-17 at 11 17 52 AM" src="https://user-images.githubusercontent.com/670099/74670486-40e61500-5177-11ea-9b73-8550f446e0b3.png">

98lenvi:
Can I take this up? I had achieved this functionality with the keyboard shortcuts, but discarded because I thought it is not an ideal behavior.

PR #262, opened 2020-03-06
[WIP] feat(schema,plugin): nested items - #241

98lenvi:
# Issue #241 
Changes for allowing nested lists (nested quotes, lists in quotes, quotes in lists will be added shortly)

### Changes
- Added `ul_list` and `ol_list` as children of `list_item` in `schema.js`
-Added a new method, `onTab` for handling creation of nested lists,
-Added support for lists in quotes, quotes in lists via `schema` changes and editing `transformBlockQuoteToList` & `transformListToBlockQuote`
### Flags
  -There is an error whenever a user presses `Tab`, **a nested list is only created when a user presses `Tab` twice**, and the nested list should be only created when the text-cursor is placed right after the list item symbol, not when it is in the middle of the list text.
  - the error generated is `Uncaught (in promise) Error: Item node should not occur outside of List nodes`
  - I will work on adding the other nesting features, then tackle the above issues.

### Related Issues
- Issue #241 
- Pull Request #250 
- Pull Request #261 (That was a mistake)

### DIFF
--- a/src/FormattingToolbar/index.js
+++ b/src/FormattingToolbar/index.js
@@ -216,22 +216,28 @@ export default class FormatToolbar extends React.Component {
     const { value } = editor;
 
     if (!lockText || pluginManager.isEditable(editor, type)) {
+      console.log(type)
       event.preventDefault();
       if (action.isClickBlockQuote(type)) {
+        console.log('action.isClickBlockQuote(type)')
         action.isSelectionList(value)
           ? action.transformListToBlockQuote(editor, type, value)
           : action.transformPtoBQSwap(editor, type);
       } else if (action.isClickHeading(type)) {
+        console.log('action.isClickHeading(type)')
         action.hasBlock(editor, type)
           ? editor.setBlocks(CONST.PARAGRAPH)
           : editor.setBlocks(type);
       } else if (action.isSelectionList(value)) {
+        console.log('action.isSelectionList(value)')
         action.currentList(value).type === type
           ? action.transformListToParagraph(editor, type)
           : action.transformListSwap(editor, type, value);
       } else if (action.isSelectionInput(value, CONST.BLOCK_QUOTE)) {
+        console.log('action.isSelectionInput(value, CONST.BLOCK_QUOTE)')
         action.transformBlockQuoteToList(editor, type);
       } else {
+        console.log('action.isSelectionInput(value, CONST.BLOCK_QUOTE)')
         action.transformParagraphToList(editor, type);
       }
     }
--- a/src/FormattingToolbar/toolbarMethods.js
+++ b/src/FormattingToolbar/toolbarMethods.js
@@ -183,8 +183,8 @@ export const currentList = value => isSelectionOLList(value) || isSelectionULLis
 export const transformListToBlockQuote = (editor, type, value) => {
   editor.withoutNormalizing(() => {
     editor
-      .unwrapBlock(CONST.LIST_ITEM)
-      .unwrapBlock(currentList(value).type)
+      //.unwrapBlock(CONST.LIST_ITEM)
+      //.unwrapBlock(currentList(value).type)
       .wrapBlock({ type, data: { tight: true } });
   });
 };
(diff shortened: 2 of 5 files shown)


irmerk:
This looks good so far!

A couple user actions are a bit buggy I think. In the demo do these two procedures:

1. Scroll to the ordered list
2. Click on "two"
3. Click block quote three times.

1. Scroll to the ordered list
2. Click on "two"
3. Click block quote once (should be a block quote in an ordered list now)
4. While selecting "two", press tab

98lenvi:
@irmerk , Thanks for pointing that one out!

DianaLease:
I think it would be best to solve the error you mentioned in "flags" before moving on. Once that error occurs, I can no longer edit any part of the document.

98lenvi:
Okay! I guess I'll have to think from a different approach to solve that.

98lenvi:
@irmerk @DianaLease. I have done a **hack** to solve the double tab issue, but I don't think it's the right approach, everything other than the nested list is working. I think for now only supporting nested quotes, lists in quotes and quotes in lists is possible from my end. 

I'll work on the nested lists once I get some more experience in slate.js, Sorry :disappointed:

98lenvi:
@irmerk @DianaLease, I'm done with the bug-fixes and also nested lists are working fine :100:. Please check it out and let me know if there are any more bugs.

irmerk:
It looks like your rebase was botched and there's at least one `<<<<<<< HEAD` remaining and it's failing the tests. Try cleaning that up for a clean build and we can look at it

Michael-Grover:
image

The indentation between different levels of nesting looks very long.

Check out Google Docs and Word for reference

image
image

98lenvi:
> It looks like your rebase was botched and there's at least one `<<<<<<< HEAD` remaining and it's failing the tests. Try cleaning that up for a clean build and we can look at it

Can you please help me with this, I have no idea how to fix it without ruining my patches.

98lenvi:
> This is coming along well. We will also need a way to outdent once in a nested list item to go back a level. Commonly, `shift + tab` is used for this. Right now in this PR, that command causes further nesting.

Okay I

> This is coming along well. We will also need a way to outdent once in a nested list item to go back a level. Commonly, `shift + tab` is used for this. Right now in this PR, that command causes further nesting.

I have added support for outdenting lists, please look into it

98lenvi:
@Michael-Grover, I think the _long identation_ might be because of the way I'm creating nested lists, 
via our plugin, the markdown for a nested list would be
- main list item 1
- - nested list item 1
  - nested list item 2
- main list item 2

Ideally, it should be 

- main list item 1
  - nested list item 1
  - nested list item 2
- main list item 2

Once I properly create lists I guess the long indentation issue should be fixed

98lenvi:
@DianaLease, I think using `SHIFT`+`TAB` isn't a good option, because when using that key combo, the following happens:
shift_tab

Michael-Grover:
> @DianaLease, I think using `SHIFT`+`TAB` isn't a good option, because when using that key combo, the following happens:
> shift_tab

This is because clicking tab in the editor changes focus to the next element instead of actually adding a tab, right? Is that a separate issue?

DianaLease:
I just tried shift + tab on the lists again in the netlify on this PR and now it is working as expected to outdent the list item! Did something change? @98lenvi

98lenvi:
@DianaLease, I didn't change anything. The behavior is only in my browser, works fine everywhere else.

98lenvi:
> > @DianaLease, I think using `SHIFT`+`TAB` isn't a good option, because when using that key combo, the following happens:
> > shift_tab
> 
> This is because clicking tab in the editor changes focus to the next element instead of actually adding a tab, right? Is that a separate issue?

@Michael-Grover, Yes I think that's the cause

irmerk:
This is great progress, but I think like a few other PRs in `markdown-editor`, we should put it on the backburner while we migrate to the latest Slate version. This will be implemented quite differently, and thus continuing work on this will only add to the complexity of migration.

dselman:
Agreed. I'm going to close this one - we will work on it for Slate 0.50 - which doesn't really have the concept of a declarative schema.

Issue #229, opened 2020-02-07
Highlighting List of Paragraphs to Turn to List


irmerk:
**Describe the bug**
Highlighting a series of paragraphs/lines and selecting list (either one) will result in a list of one with multiple lines rather than multiple units of the list.

**Screenshots**

---

<img width="169" alt="Screen Shot 2020-02-06 at 2 03 13 PM" src="https://user-images.githubusercontent.com/36460856/74056995-ffa07900-49b0-11ea-9474-8ba86b19c524.png">

---

**Additional context**
N/A

irmerk:
Still needs work, see later comments in #269

jeromesimeon:
@irmerk I believe the change in #269 will break lists containing more than one paragraph. (need to be tested, but I suspect)

PR #269, opened 2020-03-07
fix(list): converts multiple line to lists - I229

elit-altum:

# Issue #229 

### Changes
- Changed ```list.js``` to break every element into an individual ```<li>``` tag rather than all children in a single tag
Para-To-List-Fix

### Flags
- The editor list is getting converted fine, but the individual children still aren't being converted to individual list items in the markdown. Currently working to fix that.

### Related Issues
- Issue #229

### DIFF
--- a/src/plugins/list.js
+++ b/src/plugins/list.js
@@ -70,7 +70,11 @@ function List() {
       case CONST.UL_LIST:
         return <ul {...attributes}>{children}</ul>;
       case CONST.LIST_ITEM:
-        return <li {...attributes}>{children}</li>;
+        return children.map( (child) => {
+          if (child) {
+            return <li {...attributes}>{child}</li>
+          }
+        });
       default:
         return next();
     }


elit-altum:
@irmerk @DianaLease Please review !

elit-altum:
@irmerk The markdown error is due to 
```js
//toolbarMethods.js
export const transformParagraphToList = (editor, type) => {
  editor.withoutNormalizing(() => {
    editor.wrapBlock({ type, data: { tight: true } }).wrapBlock(CONST.LIST_ITEM);
  });
};
```
which is wrapping the whole content as a single ```list_item``` .
Is there anyway I can run something like a for-each loop or split the nodes into individual ```list_item``` rather than a whole?
Thank You!

elit-altum:
> @irmerk The markdown error is due to
> 
> ```js
> //toolbarMethods.js
> export const transformParagraphToList = (editor, type) => {
>   editor.withoutNormalizing(() => {
>     editor.wrapBlock({ type, data: { tight: true } }).wrapBlock(CONST.LIST_ITEM);
>   });
> };
> ```
> 
> which is wrapping the whole content as a single `list_item` .
> Is there anyway I can run something like a for-each loop or split the nodes into individual `list_item` rather than a whole?
> Thank You!
> 
> Similar is the case with
> 
> ```js
> /**
>  * A trigger to the Slate editor to swap a ul_list to ol_list or vice versa.
>  */
> export const transformListSwap = (editor, type, value) => {
>   editor.withoutNormalizing(() => {
>     editor
>       .unwrapBlock(CONST.LIST_ITEM)
>       .unwrapBlock(currentList(value).type)
>       .wrapBlock({ type, data: { tight: true } })
>       .wrapBlock(CONST.LIST_ITEM);
>   });
> };
> ```
> 
> A list made individually , which is being handled correctly by the markdown is being converted to a single `list_item` as a whole when converted from ordered to unordered list and vice versa.

I have found ```.wrapChildren(CONST.LIST_ITEM)``` to be giving the right results as expected in markdown, but we'd then have to change every List handler to work with children.

irmerk:
I'm a bit unclear base one the order of the comments. Is there a markdown error or any error? I'm not seeing it. This looks really good.

jeromesimeon:
There is definitely something wrong with the markdown. I'm suspecting the Slate DOM doesn't have the right shape:

Screenshot 2020-03-09 at 3 22 39 PM

Screenshot 2020-03-09 at 3 22 28 PM

jeromesimeon:
The Slate DOM should look as follows. I don't think this PR includes the needed changes:
```bash-3.2$ markus parse --sample test.md --slate 
3:25:39 PM - info:
{
  "object": "value",
  "document": {
    "object": "document",
    "nodes": [
      {
        "object": "block",
        "data": {
          "tight": "true"
        },
        "type": "ul_list",
        "nodes": [
          {
            "object": "block",
            "type": "list_item",
            "data": {
            },
            "nodes": [
              {
                "object": "block",
                "type": "paragraph",
                "nodes": [
                  {
                    "object": "text",
                    "text": "apples",
                    "marks": []
                  }
                ],
                "data": {
                }
              }
            ]
          }, 
          {
            "object": "block",
            "type": "list_item",
            "data": {
            },
            "nodes": [
              {
                "object": "block",
                "type": "paragraph",
                "nodes": [
                  {
                    "object": "text",
                    "text": "pear",
                    "marks": []
                  }
                ],
                "data": {
                }
              }
            ]
          }, 
          {
            "object": "block",
            "type": "list_item",
            "data": {
            },
            "nodes": [
              {
                "object": "block",
                "type": "paragraph",
                "nodes": [
                  {
                    "object": "text",
                    "text": "peaches",
                    "marks": []
                  }
                ],
                "data": {
                }
              }
            ]
          }
        ]
      }
    ],
    "data": {
    }
  }
}
```

elit-altum:
> 
> 
> There is definitely something wrong with the markdown. I'm suspecting the Slate DOM doesn't have the right shape:
> 
> Screenshot 2020-03-09 at 3 22 39 PM
> 
> Screenshot 2020-03-09 at 3 22 28 PM

Yes @jeromesimeon @irmerk This is what I was talking about. I didn't know how to proceed forward because every function on lists is using similar methods and causing similar issues.

irmerk:
I'd think there could be some check when the list function call is run, where at the beginning of the function it checks to see if there are multiple nodes selected? And if so, loop over them, like you mentioned 
> Is there anyway I can run something like a for-each loop or split the nodes into individual list_item rather than a whole?

elit-altum:
> 
> 
> For this markdown:
> 
> ```
> - apples
> - pear
> - peaches
> ```
> 
> The Slate DOM should look as follows. I don't think this PR includes the needed changes:
> 
> ```
> 3:25:39 PM - info:
> {
>   "object": "value",
>   "document": {
>     "object": "document",
>     "nodes": [
>       {
>         "object": "block",
>         "data": {
>           "tight": "true"
>         },
>         "type": "ul_list",
>         "nodes": [
>           {
>             "object": "block",
>             "type": "list_item",
>             "data": {
>             },
>             "nodes": [
>               {
>                 "object": "block",
>                 "type": "paragraph",
>                 "nodes": [
>                   {
>                     "object": "text",
>                     "text": "apples",
>                     "marks": []
>                   }
>                 ],
>                 "data": {
>                 }
>               }
>             ]
>           }, 
>           {
>             "object": "block",
>             "type": "list_item",
>             "data": {
>             },
>             "nodes": [
>               {
>                 "object": "block",
>                 "type": "paragraph",
>                 "nodes": [
>                   {
>                     "object": "text",
>                     "text": "pear",
>                     "marks": []
>                   }
>                 ],
>                 "data": {
>                 }
>               }
>             ]
>           }, 
>           {
>             "object": "block",
>             "type": "list_item",
>             "data": {
>             },
>             "nodes": [
>               {
>                 "object": "block",
>                 "type": "paragraph",
>                 "nodes": [
>                   {
>                     "object": "text",
>                     "text": "peaches",
>                     "marks": []
>                   }
>                 ],
>                 "data": {
>                 }
>               }
>             ]
>           }
>         ]
>       }
>     ],
>     "data": {
>     }
>   }
> }
> ```

Locally I can run the changes @jeromesimeon . But using them we will have to change every function handling lists. That's why I didn't proceed further.
Please see this,
Para-To-List-Fix-md

irmerk:
Also note @elit-altum that we have started the process of migrating to Slate v0.5x, which will change all these function calls. So maybe we hold off on this and wait for that migration.

elit-altum:
> 
> 
> Also note @elit-altum that we have started the process of migrating to Slate v0.5x, which will change all these function calls. So maybe we hold off on this and wait for that migration.

Sure @irmerk. Will you merge this PR ? It handles the part for the render of lists on the editor. And the linked issue is about that problem only.

PR #275, opened 2020-03-10
fix(SlateAsInputEditor): handle undo link removal - I272

### LINKED ISSUE #272: Removing link followed by Undo shortcut crashes demo
**Describe the bug**
Opening the editor and removing the templated link from popup followed by Undo shortcut (Ctrl+Z) crashes the demo.

**To Reproduce**
1. Go to `markdown-editor` demo
2. Click on the templated link 
3. Click on 'Remove' from the popup
3. Press Ctrl + Z (Undo command)
4. Demo crashes

**Expected behavior**
Preferably, the link should be restored.

**Screenshots**
link_crash

irmerk:
# Fixes #272
Resolve issue around editor breaking when undoing the removal of a link

### Changes
- Remove `async/await`
- Only use `onUndoOrRedo` if passed in

### Related Issues
- Pull Request #260


irmerk:
I suppose so. I needed to put the check `if (onUndoOrRedo) return onUndoOrRedo(editor);` to make sure this isn't called if it isn't passed in. Then after that, there were still crashing errors if `await` was present.

jeromesimeon:
> I suppose so. I needed to put the check `if (onUndoOrRedo) return onUndoOrRedo(editor);` to make sure this isn't called if it isn't passed in. Then after that, there were still crashing errors if `await` was present.

I'll take your word for it. I just think it would be useful to understand why.

irmerk:
@adriaan-pelzer this is the problem I had mentioned to you. If you or anyone else can provide insight as to why this was an issue, 👍

jeromesimeon:
> @adriaan-pelzer this is the problem I had mentioned to you. If you or anyone else can provide insight as to why this was an issue, 👍

What does the Slate documentation say about undo/redo. Are those async calls? Are there any Slate samples out there to show how those calls are used. Have you tested the undo/redo functionality with this new code?

irmerk:
The [docs](https://docs.slatejs.org/v/v0.47/slate-core/commands#history-commands) regarding this aren't very helpful, but they aren't really helpful for most things. 

[This](https://github.com/ianstormtaylor/slate/blob/v0.47/examples/history/index.js#L93-L97) is the closest to an example of using undo in their `0.47` branch.

Yes the functionality looks like it works.

dselman:
I don't understand why we need this - or what the prop passed in should do to the editor...

Issue #280, opened 2019-11-06
Hitting enter after a heading inserts blank line


dselman:
**Describe the bug**
Hitting enter after a heading inserts blank line

**To Reproduce**
1. Position cursor at end of "Heading One" in the editor
2. Hit enter
3. Note that an extra blank line has been inserted

**Expected behavior**
Enter after headings should not insert an extra blank line.

PR #281, opened 2020-03-12
fix(SlateAsInputEditor): removed blank line on hitting enter after heading - I280

elit-altum:

# Issue #280 
Removed the extra paragraph block which caused the spacing by inserting an empty text string

### Changes
- Now an empty text string is added on hitting enter instead of a paragraph which also allows multiple line headings
- A backspace on the cursor before typing the heading on the new line converts it back into a paragraph.

### Flags
- Not sure of the solution implemented would need a review

### Related Issues
- Issue #280

### DIFF
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -327,7 +327,7 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
 
     // when you hit enter after a heading we insert a paragraph
     event.preventDefault();
-    editor.insertBlock(CONST.PARAGRAPH);
+    editor.insertText('');
     return next();
   };
 
--- a/src/SlateAsInputEditor/index.js
+++ b/src/SlateAsInputEditor/index.js
@@ -319,15 +319,14 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
       return false;
     }
 
-    // if you hit enter inside anything that is not a heading
-    // we use the default behavior
-    if (!startBlock.type.startsWith('heading')) {
-      return next();
+    // when you hit enter after a heading we insert a paragraph
+    if (startBlock.type.startsWith('heading')) {
+      event.preventDefault();
+      return editor.insertBlock(CONST.PARAGRAPH);
     }
 
-    // when you hit enter after a heading we insert a paragraph
-    event.preventDefault();
-    editor.insertText('');
+    // if you hit enter inside anything that is not a heading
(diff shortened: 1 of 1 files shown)


elit-altum:
@dselman @DianaLease I have tried to implement a solution. Please review!

irmerk:
This looks great to me. @Michael-Grover is this behavior what we'd want?

Michael-Grover:
> 2\. When a user hits enter after a heading, do we want the user to be in a new heading on the next line? In Google docs, the next line becomes a paragraph, not another heading.

I agree with @DianaLease . We should do what Microsoft Word and Google Docs do. I made a video response explaining in more detail, but basically I think when a user presses enter on a header line, the next line should be normal text. Let me know if I can clarify anything from the video @elit-altum 

https://www.loom.com/share/71dfef184fe6451dacdc4209faa0cee4

elit-altum:
> 
> 
> > This looks great to me. @Michael-Grover is this behavior what we'd want?
> 
> > 2. When a user hits enter after a heading, do we want the user to be in a new heading on the next line? In Google docs, the next line becomes a paragraph, not another heading.
> 
> I agree with @DianaLease . We should do what Microsoft Word and Google Docs do. I made a video response explaining in more detail, but basically I think when a user presses enter on a header line, the next line should be normal text. Let me know if I can clarify anything from the video @elit-altum
> 
> https://www.loom.com/share/71dfef184fe6451dacdc4209faa0cee4

Thank you for that comprehensive information @Michael-Grover it helped a lot! I'll get back to it. Just one thing, on a backspace the text-type was getting converted back to paragraph which was a smaller font that's why the cursor was changing sizes

src/SlateAsInputEditor/index.js
```diff
@@ -319,15 +319,14 @@ const SlateAsInputEditor = React.forwardRef((props, ref) => {
...
+      event.preventDefault();
+      return editor.insertBlock(CONST.PARAGRAPH);
     }
 
-    // when you hit enter after a heading we insert a paragraph
-    event.preventDefault();
-    editor.insertBlock(CONST.PARAGRAPH);
+    // if you hit enter inside anything that is not a heading
+    // we use the default behavior
     return next();
```
elit-altum: This ```next()``` after the creation of the paragraph was adding that extra line. :)

elit-altum:
@Michael-Grover @irmerk @DianaLease Please review !
I think it's fixed

PR #284, opened 2020-03-12
feat(LinkPopover): copy link href to clipboard

elit-altum:

# Issue #213 
Continues on https://github.com/accordproject/markdown-editor/pull/246 and adds a feature to copy the link text to the clipboard by providing a 'copy' button.
Only displayed while editing the link.

### Changes
- Adds a new button to copy the link text to the clipboard.

### ScreenShot
copy-linkText-to-Clipboard

### Flags
- The UI looks bad. Really need @Michael-Grover's help to fix the layout, as three buttons will take lots of space. (Maybe use icons only?)
- We need to alert the user that the link is copied maybe through a tooltip or a ```Popup``` from Semantic-UI. Need suggestions about it

### Related Issues
- Pull Request #246


elit-altum:
@irmerk @DianaLease @Michael-Grover Please review!

irmerk:
@Michael-Grover any thoughts on this?

Michael-Grover:
I will work on an improved hyperlink UI and share that when it's ready

elit-altum:
@Michael-Grover and @irmerk I have implemented the designs. Please review.
Some points:
1. The ```Button``` from Semantic-UI allows very less modification therefore the button colors are not changed. Should I remove and write the CSS for a button from scratch ?
2. Addition of tooltips would require the addition of state as well. A workaround could be found if we use the ```hover``` property in a different CSS file. Currently, the ```<title>``` of the SVG's is providing a tootlip. Please tell me if it's good enough, if not, which approach should I prefer?

Michael-Grover:
@DianaLease may be able to help with point 1 as she has made some similar buttons in a different product

DianaLease:
> 1. The `Button` from Semantic-UI allows very less modification therefore the hover button colors are not changed. Should I remove and write the CSS for a button from scratch ?

I would suggest either creating a new `styled-component` for buttons instead of using the semantic ui button, _OR_ extending the semantic-ui button with styled components using the `styled(Button)` syntax, so something like...
```
import { Button } from semantic-ui-react;
const StyledButton = styled(Button)`
  &:hover {
    // add your styles here
  }
`;
```

elit-altum:
> > 1. The `Button` from Semantic-UI allows very less modification therefore the hover button colors are not changed. Should I remove and write the CSS for a button from scratch ?
> 
> I would suggest either creating a new `styled-component` for buttons instead of using the semantic ui button, _OR_ extending the semantic-ui button with styled components using the `styled(Button)` syntax, so something like...
> 
> ```
> import { Button } from semantic-ui-react;
> const StyledButton = styled(Button)`
>   &:hover {
>     // add your styles here
>   }
> `;
> ```

@DianaLease I have tried this approach. It's not working. I'll make a new styled div. Thank you for the help!

elit-altum:
Closing this to move to development branch.

Issue #334, opened 2020-04-13
Rich Text Editor Overlaps Markdown Text Editor


shrey-sachdeva2000:
**Describe the bug**
When the `TextArea` of `MarkdownTextEditor` is resized to increase its width, the `RichTextEditor` overlaps the `TextArea` resize button and the `MarkdownTextEditor`  can't be resized again.

**To Reproduce**
1. Go to the [demo](https://accordproject-markdown-editor.netlify.com/)
2. Click on the `Markdown Text Editor` resize button
3. Increase its width
4. The `Markdown Text Editor` can't be resized again

**Expected behavior**
Either the `MarkdownTextEditor` should have a fixed width so the contents of both the Editors
don't overlap _or_ the `MarkdownTextEditor` should still be resizable.

**Screenshots**
![Markdown Editor (4)](https://user-images.githubusercontent.com/47667325/79148800-bff87200-7de3-11ea-94d1-5f3f09b90e49.gif)

**Additional Context**
I'd like to work on this!

Michael-Grover:
Would it be difficult to make the border between the two editors movable, instead of dragging on the edge of one editor just changing the size of one editor but not the other? For example if I move the border right to expand the editor on the left, the editor on the right will automatically become smaller to accommodate. Like how you can move the borders between editors in codepen

https://codepen.io/msdr/pen/LYYPqpa

shrey-sachdeva2000:
Thats sounds great! I created a prototype using [React Resizable](https://www.npmjs.com/package/react-resizable). It looks like this:

![Markdown Editor (8)](https://user-images.githubusercontent.com/47667325/79218434-08f50880-7e6e-11ea-81f0-cf9ed408f53a.gif)

Although, I think its better to not keep the `RichTextEditor` resizable because our `ToolbarMenu` would go out of proportion and snap. Something like this: 

![Markdown Editor (6)](https://user-images.githubusercontent.com/47667325/79217302-422c7900-7e6c-11ea-9f67-45a0a507c752.gif)

So, only the `MarkdownEditor` would be resizable and its max-width would be set to its original width so it doesn't collapse the `RichTextEditor`. 

![Markdown Editor (7)](https://user-images.githubusercontent.com/47667325/79217202-114c4400-7e6c-11ea-9566-9bf34a067bee.gif)


Please let me know your thoughts on this! @Michael-Grover @DianaLease @irmerk

Michael-Grover:
[...]

PR #315, opened 2020-04-11
feat(ToolbarMenu): sticky formatting toolbar position - I314

shrey-sachdeva2000:
Signed-off-by: Shrey Sachdeva <shreysachdeva.2000@gmail.com>

# Issue #314 
Make `ToolbarMenu` sticky

### Changes
- `ToolbarMenu` is now sticky and attaches to the top of the editor on scroll

### Flags
- N/A

PR #338, opened 2020-04-13
fix(styles): fix markdown editor width - I334

shrey-sachdeva2000:
Signed-off-by: Shrey Sachdeva <shreysachdeva.2000@gmail.com>

# Issue #334 
Fix `Markdown Editor` width so it doesn't overlap with `Rich Text Editor` on resizing.

### Changes
- Set a `max-width` on `Markdown Editor`

### Flags
- N/A

### DIFF
--- a/src/styles.css
+++ b/src/styles.css
@@ -17,7 +17,7 @@
     height: 100%;
     width: 100%;
     border: 0;
-    max-width: 45vw;
+    max-width: 100%;
 }
 
 .ap-markdown-editor .error {
--- a/src/styles.css
+++ b/src/styles.css
@@ -17,7 +17,7 @@
     height: 100%;
     width: 100%;
     border: 0;
-    max-width: 100%;
+    resize: none;
 }
 
 .ap-markdown-editor .error {
--- a/src/styles.css
+++ b/src/styles.css
@@ -17,6 +17,7 @@
     height: 100%;
     width: 100%;
     border: 0;
+    max-width: 45vw;
 }
 
 .ap-markdown-editor .error {


irmerk:
Note to reviewers: Will need to test within `cicero-ui` as well. Related to #315 so review in conjunction.

irmerk:
@shrey-sachdeva2000 can this just be forced in within the demo?

DianaLease:
> @shrey-sachdeva2000 can this just be forced in within the demo?

I agree with this ^ The component itself should not limit its width because others using it may not want that limitation. Let's just style it within the context of the demo.

shrey-sachdeva2000:
Sure! That makes sense!

shrey-sachdeva2000:
If we set `max-width: 100%` instead of `max-width: 45vw` the textarea would stay within its parent element. Would that be a good option?

DianaLease:
> If we set `max-width: 100%` instead of `max-width: 45vw` the textarea would stay within its parent element. Would that be a good option?

Yes, that sounds good to me!

DianaLease:
> If we set `max-width: 100%` instead of `max-width: 45vw` the textarea would stay within its parent element. Would that be a good option?

Sorry, looking at this code again. Setting `max-width` is kind of redundant since the `width` is already set to `100%`. What is the issue exactly? When I resize the page in the currently deployed demo, the two components never overlap in the demo.

shrey-sachdeva2000:
> Sorry, looking at this code again. Setting `max-width` is kind of redundant since the `width` is already set to `100%`. What is the issue exactly? When I resize the page in the currently deployed demo, the two components never overlap in the demo.

It is actually after one tries to resize the `MarkdownEditor` textarea that it flows out of its set width and it's not resizable again. Not sure if this happens to be an issue with a specific browser itself. Please check the screenshot here: https://github.com/accordproject/markdown-editor/issues/334

irmerk:
Yeah agree with @DianaLease, make this all static and get rid of resizing and such. It's a demo and should just be static.

irmerk:
Looks good to me, just rebase to be up to date with `master`

shrey-sachdeva2000:
> Looks good to me, just rebase to be up to date with `master`

Sure, the PR is now ready to merge.

irmerk:
You'll want to get rid of fc5c50b, we don't want merge commits in the git log. I think always lean towards rebasing.

shrey-sachdeva2000:
Ohh my bad. It's fixed now!

PR #346, opened 2020-04-14
fix(HyperlinkModal.js): prevent insertion of links if URL is empty

algomaster99:
Signed-off-by: Aman Sharma <mannu.poski10@gmail.com>

### Changes
- Add a state `blankUrlError` to keep track of the URL if valid data is entered in the modal.
- If the URL is blank, error is displayed something like this:
Screenshot from 2020-04-14 12-43-35
- For further demonstration, here is a screencast:
accord

### Flags
- Instead of checking if the URL is empty, we should add RegEx to validate if it is a URL or not.

### DIFF
--- a/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
+++ b/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
@@ -96,7 +96,7 @@ const HyperlinkModal = React.forwardRef(({ ...props }, ref) => {
     // If the form is just opened, focus the Url input field
     if (props.showLinkModal) {
       setOriginalSelection(editor.selection);
-      setApplyStatus(refHyperlinkTextInput.current.props.defaultValue);
+      setApplyStatus(!!refHyperlinkTextInput.current.props.defaultValue);
       refHyperlinkTextInput.current.focus();
     }
   }, [editor, props.showLinkModal]);
@@ -118,7 +118,7 @@ const HyperlinkModal = React.forwardRef(({ ...props }, ref) => {
 
   const handleUrlInput = (event) => {
     Transforms.select(editor, originalSelection);
-    event.target.value ? setApplyStatus(true) : setApplyStatus(false);
+    setApplyStatus(!!event.target.value);
   }
 
   return (
--- a/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
+++ b/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
@@ -8,7 +8,7 @@ import {
 } from 'slate';
 import styled from 'styled-components';
 import {
-  Button, Form, Input,
+  Button, Form, Input, Label,
 } from 'semantic-ui-react';
 
 import { insertLink, isSelectionLink, unwrapLink } from '../plugins/withLinks';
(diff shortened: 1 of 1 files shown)


Michael-Grover:
I don't think this is necessary, what is the argument for adding it?

algomaster99:
@Michael-Grover If someone accidentally puts an empty link, and didn't undo, I just thought it would be difficult for them to delete those extra characters since it can't be located.

DianaLease:
> I don't think this is necessary, what is the argument for adding it?

It doesn't make sense for a user to create a link with no URL. We should prevent that I think.

DianaLease:
> > I don't think this is necessary, what is the argument for adding it?
> 
> It doesn't make sense for a user to create a link with no URL. We should prevent that I think.

Or maybe we could just disable the "Apply" button if URL field is blank? @Michael-Grover @algomaster99

algomaster99:
@DianaLease That also makes sense. But I think flashing an error is more intuitive to the users.

Michael-Grover:
I think disabling the "Apply" button like Diana suggested is sufficient

algomaster99:
@Michael-Grover @DianaLease Sure, I will make the required changes.

jeromesimeon:
@algomaster99 Please make sure the change still allows the component to _render_ an empty link. If the user loads a contract or template written directly in markdown (with a link with an empty URL which is allowed) into the editor it should still work.

algomaster99:
@jeromesimeon my changes will **prevent** the user to add an empty link. I can revert these changes in that case but before I do it, can you tell me such use case when an empty URL can be rendered?

jeromesimeon:
> @jeromesimeon my changes will **prevent** the user to add an empty link. I can revert these changes in that case but before I do it, can you tell me such use case when an empty URL can be rendered?

I understand. I'm saying: what if the link is not added through the UI (but pre-exists in the underlying markdown).

algomaster99:
@jeromesimeon Okay, that makes sense. I will test that also before submitting the changes.

DianaLease:
@jeromesimeon @algomaster99 Your changes shouldn't affect rendering, but you can double check by adding an empty link to the markdown in the demo code (ie `[empty link]()`) and making sure it still renders.

algomaster99:
@DianaLease sure, thanks for the pointer :)

src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
```diff
@@ -95,6 +96,7 @@ const HyperlinkModal = React.forwardRef(({ ...props }, ref) => {
     // If the form is just opened, focus the Url input field
     if (props.showLinkModal) {
       setOriginalSelection(editor.selection);
+      setApplyStatus(refHyperlinkTextInput.current.props.defaultValue);
```
DianaLease, line 99: let's use a boolean instead of the string, so you can instead do `setApplyStatus(!!refHyperlinkTextInput.current.props.defaultValue)`

src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
```diff
@@ -114,6 +116,11 @@ const HyperlinkModal = React.forwardRef(({ ...props }, ref) => {
     props.setShowLinkModal(false);
   };
 
+  const handleUrlInput = (event) => {
+    Transforms.select(editor, originalSelection);
+    event.target.value ? setApplyStatus(true) : setApplyStatus(false);
```
DianaLease, line 121: here you can use the same syntax to simplify by making this line just `setApplyStatus(!!event.target.value)`

irmerk:
Looks good! Need to rebase again to be up to date with `master` and we can get this merged in.

Issue #335, opened 2020-04-13
Lists Break on Enter


shrey-sachdeva2000:
**Describe the bug**
Pressing <kbd>Enter</kbd> on an item in the List doesn't create a new item in the List.

**To Reproduce**
1. Go to the [demo](https://accordproject-markdown-editor.netlify.com/)
2. Click on a List item
3. Press <kbd>Enter</kbd>
4. A new line is created instead of List item

**Expected behavior**
On pressing <kbd>Enter</kbd>, a new list item should be created. While, on pressing <kbd>Shift+Enter</kbd> a new line within the same item should be created.

**Screenshots**
![lists](https://user-images.githubusercontent.com/47667325/79151713-8d04ad00-7de8-11ea-9c80-9b94008ef464.PNG)

Michael-Grover:
I want to add to this that pressing enter once should create a new list item, but pressing enter twice (or once on the new empty list item) should remove the list item and exit the list. See how Google Docs works

irmerk:
Opened an issue within Slate tracking this [here](https://github.com/ianstormtaylor/slate/issues/3624).

Issue #323, opened 2020-04-13
Roundtrip Ordered List


irmerk:
**Describe the bug**
A user cannot roundtrip the ordered list in the demo

**To Reproduce**
1. Go to the [live demo](https://accordproject-markdown-editor.netlify.com/)
2. Highlight the entire ordered list (one, two, three)
3. Deselect numbered list in the format toolbar
4. See they have turned into an incorrect bulleted list
5. Format back into a numbered list with the format toolbar
6. See further error rendering the numbers in the list

**Expected behavior**
Roundtripping lists should behave as expected.

**Screenshots**
<img width="123" alt="Screen Shot 2020-04-13 at 10 03 34 AM" src="https://user-images.githubusercontent.com/36460856/79126527-1c20af00-7d6e-11ea-981f-c7d98ad2c62a.png">

irmerk:
Note unwrapping the existing ordered list makes them unordered list items in the markdown display on the left. So selecting numbered list again makes it:
```md
1. - one
2. - two
3. - three
```

PR #350, opened 2020-04-14
feat(RichTextEditor): break from empty block on enter - I321

### LINKED ISSUE #321: Cannot Escape Blockquote
**Describe the bug**
A user has difficulty escaping block quotes

**To Reproduce**
1. Go to the [live demo](https://accordproject-markdown-editor.netlify.com/)
2. Click within the block quote "`This is a quote.`"
3. Press `Enter` as many times as you wish
4. Does not escape blockquote

**Expected behavior**
Pressing enter twice should escape the blockquote.

irmerk:
# Closes #321 
Pressing enter twice should escape a block(quote)

### Changes
- Break from any current block type (UL, OL, Blockquote)
  - If the current selection is an empty text node
- Correct the passing of `children` into elements

### Flags
- Breaking from a list results in similar bugs to #323 and #335

### Related Issues
- Issue #323
- Issue #335

### DIFF
--- a/src/RichTextEditor/index.js
+++ b/src/RichTextEditor/index.js
@@ -5,13 +5,13 @@ import { CiceroMarkTransformer } from '@accordproject/markdown-cicero';
 import { HtmlTransformer } from '@accordproject/markdown-html';
 import { SlateTransformer } from '@accordproject/markdown-slate';
 import isHotkey from 'is-hotkey';
-import {
-  Editable, withReact, Slate
-} from 'slate-react';
-import { Node, createEditor } from 'slate';
+import { Editable, withReact, Slate } from 'slate-react';
+import { Node, createEditor, Transforms } from 'slate';
 import { withHistory } from 'slate-history';
 import PropTypes from 'prop-types';
-import HOTKEYS, { formattingHotKeys } from './utilities/hotkeys';
+import HOTKEYS, {
+  ENTER, ENTER_BREAKS, formattingHotKeys
+} from './utilities/hotkeys';
 import withSchema from './utilities/schema';
 import Element from './components';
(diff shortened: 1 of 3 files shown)


irmerk:
I think the error is because `Node.ancestors` is still running after I do the `Transforms.unwrapNodes`

src/RichTextEditor/index.js
```diff
@@ -58,6 +58,19 @@ const RichTextEditor = (props) => {
   const { isEditable, canBeFormatted } = props;
 
   const onKeyDown = useCallback((event) => {
+    if (isHotkey(ENTER, event)) {
+      // eslint-disable-next-line no-restricted-syntax
+      for (const [n] of Node.ancestors(editor, editor.selection.focus.path, { reverse: true })) {
```
DianaLease, line 63: To add to our discussion we had over zoom - I think we don't want to pass the reverse option so that we go bottom up as the docs say: "By default, the order is bottom-up, from lowest to highest ancestor in the tree."
Screen Shot 2020-04-14 at 4 41 57 PM

irmerk, line 63: @DianaLease this goes from the bottom up in the tree: https://docs.slatejs.org/api/nodes#static-methods

irmerk: When I pass reverse in, I get this log order:
1. paragraph
2. block quote
3. editor

When I don't pass reverse in, I get this log order:
1. editor
2. block quote
3. paragraph

irmerk: Could be semantics and they are saying the editor is at the bottom?

DianaLease: wow, that is confusing!

PR #351, opened 2020-04-15
feat(HyperlinkModal): changed the UI and added icons - I213-293

elit-altum:
Signed-off-by: elit-altum <manan.sharma311@gmail.com>

# Issue #213 #293 
Continues on https://github.com/accordproject/markdown-editor/pull/246 and adds a feature to copy the link text to the clipboard by providing a 'copy' button.
Only displayed while editing the link.

### Changes
- Adds a new button to copy the link text to the clipboard.
- Modifies the current Hyperlink-UI

### ScreenShot
image

### Related Issues
- Pull Request #246 
- Pull Request #284

PR #353, opened 2020-04-16
feat(HyperlinkModal): add hyperlink on pressing enter

algomaster99:
Signed-off-by: Aman Sharma <mannu.poski10@gmail.com>

### Changes
- Fill hyperlink modal and press <kbd>Enter</kbd> to apply the link.

Here is an example demonstration:
accord

### Flags
- I have used `document.getElementById('form')` to get the form HTML node. But I think it is not the correct way to do it in React. I tried using `useRef()` / `React.createRef()` but I got an object instead of the HTML node.
- The reason I need an HTML node is to access the URL and text using `.url.value` and `text.value`.

### DIFF
--- a/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
+++ b/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
@@ -3,6 +3,7 @@ import React, {
 } from 'react';
 import PropTypes from 'prop-types';
 import { ReactEditor, useEditor } from 'slate-react';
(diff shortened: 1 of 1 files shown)


irmerk:
This will need to rebase changes from #351

algomaster99:
@irmerk done

algomaster99:
@DianaLease yes, you are right. This is already a feature on master. Sorry for not noticing. :sweat_smile:

PR #357, opened 2020-04-16
feat(HyperlinkModal): added tooltips to action icons

elit-altum:

### Changes
- Added tooltips to the HyperlinkModal action icons using Semantic-UI Popup

Screenshot:
New-Tooltips

### Flags
- Please review the tooltip messages. The current ones are:
   - Copy Icon: Copy hyperlink text
   - Remove Icon: Remove hyperlink
   - Open Icon: Open link in a new tab

### Related Issues
- Pull Request #351

### DIFF
--- a/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
+++ b/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
@@ -241,8 +241,7 @@ const HyperlinkModal = React.forwardRef(({ ...props }, ref) => {
               <Popup
                 trigger={
                   <LinkIconHolder
-                    disabled={!isSelectionLink(editor)}
-                    onClick={removeLink}
+                    onClick={isSelectionLink(editor) ? removeLink : null}
                     aria-label="Remove hyperlink"
                   >
                     <DeleteIcon />
--- a/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
+++ b/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
@@ -8,7 +8,7 @@ import {
 } from 'slate';
 import styled from 'styled-components';
 import {
-  Button, Form, Input,
+  Button, Form, Input, Popup
 } from 'semantic-ui-react';
 
 import { insertLink, isSelectionLink, unwrapLink } from '../plugins/withLinks';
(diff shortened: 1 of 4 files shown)


elit-altum:
@irmerk @DianaLease @Michael-Grover Please have a look!

Michael-Grover:
Can we remove the fade-in animation when this modal loads, and make the triangle at the top of the modal look like the following, instead of being filled in gray? 

image

Sorry for not catching this in the last PR. Other than that it looks great!

elit-altum:
> 
> 
> Can we remove the fade-in animation when this modal loads, and make the triangle at the top of the modal look like the following, instead of being filled in gray?
> 
> image
> 
> Sorry for not catching this in the last PR. Other than that it looks great!

@Michael-Grover I don't think my PR has caused this. This might be due to some built-in Semantic styling. I'll take a look.

src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
```diff
@@ -219,25 +224,49 @@ const HyperlinkModal = React.forwardRef(({ ...props }, ref) => {
...
+                }
+                content="Copy hyperlink text"
+                inverted
+                position='bottom left'
+                style={popupStyles}
+              />
+              <Popup
+                trigger={
+                  <LinkIconHolder
+                    disabled={!isSelectionLink(editor)}
```
irmerk, line 244: Looks like this already existed from the previous PR, but what does `disabled={!isSelectionLink(editor)}` do here? I don't see any UI difference.

irmerk:
> @Michael-Grover I don't think my PR has caused this. This might be due to some built-in Semantic styling. I'll take a look.

Yeah we've run into issues like this with Semantic UI React before. This shouldn't be a blocker to this PR and can be a separate issue to work on in the future.

src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
elit-altum: @irmerk this was used when a button was being used for the 'remove' option. As we are not using it now we can remove this.
Also can I use
```js 
onClick={isSelectionLink(editor) ? removeLink : null}
```
on the delete iconHolder so that it can't be used if the selection is not already a link i.e. one would not be able to use this while creating a link ?

irmerk: Yeah I think that would be good to do.

elit-altum: @irmerk Just made the changes. Please review!
Thank You

PR #358, opened 2020-04-16
chore(dep) Upgrade to latest version of markdown transform

jeromesimeon:
Signed-off-by: Jerome Simeon <jeromesimeon@me.com>

### Changes
- Upgrade to latest markdown transform
- Slate DOM has bold/italics annotations on parent node in some cases (instead of text node)
- Tested on demo for non-clauses
- Includes cloning Slate DOM when transforming to markdown to avoid immutability issues
- Includes fixes for https://github.com/accordproject/cicero-ui/issues/378 by @irmerk 
- Removes now unnecessary cloning in `handleCopyOrCut`

### Flags
- Done minimal local testing

### DIFF
--- a/src/RichTextEditor/index.js
+++ b/src/RichTextEditor/index.js
@@ -100,7 +100,7 @@ const RichTextEditor = (props) => {
     // https://github.com/ianstormtaylor/slate/issues/3577
     // We need to take a functional approach
     // https://github.com/accordproject/markdown-transform/issues/203
-    const SLATE_CHILDREN = JSON.parse(JSON.stringify(Node.fragment(editor, editor.selection)));
+    const SLATE_CHILDREN = Node.fragment(editor, editor.selection);
     const SLATE_DOM = {
       object: 'value',
       document: {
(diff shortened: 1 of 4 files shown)


src/RichTextEditor/index.js
```diff
@@ -100,7 +100,7 @@ const RichTextEditor = (props) => {
     // https://github.com/ianstormtaylor/slate/issues/3577
     // We need to take a functional approach
     // https://github.com/accordproject/markdown-transform/issues/203
-    const SLATE_CHILDREN = JSON.parse(JSON.stringify(Node.fragment(editor, editor.selection)));
+    const SLATE_CHILDREN = Node.fragment(editor, editor.selection));
```
DianaLease, line 103: I think this has an extra paren at the end

jeromesimeon:
> handleCopyOrCut

Good catch. Did some local testing and it seems to work fine (with the cloning removed).

src/RichTextEditor/index.js
jeromesimeon: I really thought I fixed that faster than you could notice!

PR #370, opened 2020-04-21
feat(RichTextEditor): handle breaking from list - I335

### LINKED ISSUE #335: Lists Break on Enter

irmerk:
# Closes #335
Handle adding new list items

### Changes
- New HOF to handle list logic
- On enter, if in a `list_item`, create and insert a new `list_item`

### Flags
- N/A

### Related Issues
- Issue #321
- Pull Request #350

### DIFF
--- a/src/RichTextEditor/index.js
+++ b/src/RichTextEditor/index.js
@@ -6,19 +6,18 @@ import { HtmlTransformer } from '@accordproject/markdown-html';
 import { SlateTransformer } from '@accordproject/markdown-slate';
 import isHotkey from 'is-hotkey';
 import { Editable, withReact, Slate } from 'slate-react';
-import { Node, createEditor, Transforms } from 'slate';
+import { Node, createEditor } from 'slate';
 import { withHistory } from 'slate-history';
 import PropTypes from 'prop-types';
-import HOTKEYS, {
-  ENTER, ENTER_BREAKS, ENTER_DOUBLE_BREAKS, formattingHotKeys
-} from './utilities/hotkeys';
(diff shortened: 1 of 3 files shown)


Michael-Grover:
https://www.loom.com/share/fed85a7d07e441db8a8e7ef145f20cc9

It looks like quote blocks can't be escaped. If I press enter in a quote block it creates another quote block, and if I press enter again, my cursor just moves back into the previous quote block, creating a loop. Does this need another issue? Do you know if we've already defined expected behavior for this?

@irmerk

irmerk:
Yeah open a new issue with `Slate57` label @Michael-Grover

shrey-sachdeva2000:
I think there is also no mechanism currently to enter a new line in the same list item. Google Docs and other editors handle this with `Shift+Enter`. Should I open a new issue for this? @irmerk

Michael-Grover:
@shrey-sachdeva2000 Yes I think this makes sense. That is how Microsoft Word handles the issue as well

irmerk:
Agreed, yes that should be a new Issue.

PR #371, opened 2020-04-21
fix(utilities): handle block toggles - I323

### LINKED ISSUE #323: Roundtrip Ordered List

irmerk:
# Closes #323
Handle roundtrip action of toggling block types

### Changes
- Unwraps all block types
- If the selection began as not a block, wrap in the format
  - If the format is a list, wrap each child node in a `list_item`

### Flags
- N/A

### DIFF
--- a/src/RichTextEditor/utilities/toolbarHelpers.js
+++ b/src/RichTextEditor/utilities/toolbarHelpers.js
@@ -1,5 +1,7 @@
-import { Editor, Transforms } from 'slate';
-import * as SCHEMA from './schema';
+import { Node, Editor, Transforms } from 'slate';
+import {
+  LIST_ITEM, BLOCK_QUOTE, LIST_TYPES, PARAGRAPH
+} from './schema';
 
 export const isBlockActive = (editor, format) => {
   const [match] = Editor.nodes(editor, {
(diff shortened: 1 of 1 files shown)


Michael-Grover:
Does this have any effect on the functionality from a users' perspective? I can't tell

irmerk:
@Michael-Grover go to the [current demo](https://accordproject-markdown-editor.netlify.app/), highlight a list, toggle it on and off.

src/RichTextEditor/utilities/toolbarHelpers.js
```diff
@@ -16,25 +18,31 @@ export const isMarkActive = (editor, format) => {
...
+  const isQuote = input => input === BLOCK_QUOTE;
+  const isListItem = input => input === LIST_ITEM;
+
+  /* Clear selection of block types */
+  Transforms.unwrapNodes(editor, { match: n => isQuote(n.type), split: true });
+  Transforms.unwrapNodes(editor, { match: n => isListItem(n.type), split: true });
+  Transforms.unwrapNodes(editor, { match: n => LIST_TYPES.includes(n.type), split: true });
+
+  if (!isActive) {
+    const formattedBlock = { type: format, children: [], data: { tight: true } };
```
DianaLease, line 31: Don't we only need the `tight` property if it's a list? I don't _think_ we need it if it's a block quote.

irmerk: I didn't see a change when doing so, but I made this be an empty object if it's a block quote.

PR #377, opened 2020-04-23
feat(RichTextEditor): add ID to headings

DianaLease:

- Adds `id` attributes to headings. The function for transforming the heading text into a viable anchor was taken from here: https://github.com/thlorenz/anchor-markdown-header/blob/87e4cca4618271e363f4af9697df0f73f23b3d3f/anchor-markdown-header.js#L20

### DIFF
--- a/src/RichTextEditor/components/index.js
+++ b/src/RichTextEditor/components/index.js
@@ -20,14 +20,27 @@ const Element = (props) => {
     attributes, children, element, customElements
   } = props;
   const { type, data } = element;
+  const headings = [H1, H2, H3, H4, H5, H6];
+  let headingId;
+  if (headings.includes(type)) {
+    // https://github.com/thlorenz/anchor-markdown-header/blob/87e4cca4618271e363f4af9697df0f73f23b3d3f/anchor-markdown-header.js#L20
+    headingId = element.children[0].text.replace(/ /g,'-')
+    // escape codes
+    .replace(/%([abcdef]|\d){2,2}/ig, '')
+    // single chars that are removed
+    .replace(/[\/?!:\[\]`.,()*"';{}+=<>~\$|#@&–—]/g,'')
+    // CJK punctuations that are removed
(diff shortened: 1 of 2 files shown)


DianaLease:
> * Should we have headers for Clauses too?

Right now clause nodes are given an `id` which is the clause id, which we can use to navigate to them with the Navigation component. Works for now, but we could rethink this approach!

jeromesimeon:
> > * Should we have headers for Clauses too?
> 
> Right now clause nodes are given an `id` which is the clause id, which we can use to navigate to them with the Navigation component. Works for now, but we could rethink this approach!

Right. that makes sense actually. I have others questions about clause ids, but that's totally ok as separate issue.

src/RichTextEditor/components/index.js
```diff
@@ -20,14 +20,27 @@ const Element = (props) => {
     attributes, children, element, customElements
   } = props;
   const { type, data } = element;
+  const headings = [H1, H2, H3, H4, H5, H6];
```
irmerk, line 23: `HEADINGS` could be pulled from `src/RichTextEditor/utilities/schema.js`

src/RichTextEditor/components/index.js
```diff
@@ -20,14 +20,27 @@ const Element = (props) => {
...
+    // https://github.com/thlorenz/anchor-markdown-header/blob/87e4cca4618271e363f4af9697df0f73f23b3d3f/anchor-markdown-header.js#L20
+    headingId = element.children[0].text.replace(/ /g,'-')
+    // escape codes
+    .replace(/%([abcdef]|\d){2,2}/ig, '')
+    // single chars that are removed
+    .replace(/[\/?!:\[\]`.,()*"';{}+=<>~\$|#@&–—]/g,'')
+    // CJK punctuations that are removed
+    .replace(/[。？！，、；：“”【】（）〔〕［］﹃﹄“ ”‘’﹁﹂—…－～《》〈〉「」]/g, '')
+    ;
+  }
```
irmerk, line 35: Could we make the `headingId` creation a utility function and declare `headingId` in a ternary?

Issue #376, opened 2020-04-23
Clicking on Pasted Link Changes Viewport


shrey-sachdeva2000:
**Describe the bug**
Pasting a link below the inital viewport of the editor and clicking on it, brings the editor viewport to the top. So the user has to scroll all the way down to edit the link.

**To Reproduce**
1. Go to [the demo](https://accordproject-markdown-editor.netlify.app/)
2. Copy a link
3. Scroll down and paste it 
4. Click on the link
5. See the bug

**Expected behavior**
The viewport shouldn't change on clicking the link.

**Screenshots**
![Markdown Editor (9)](https://user-images.githubusercontent.com/47667325/80115743-bcfb3f80-85a2-11ea-8014-6749c28717c1.gif)

PR #380, opened 2020-04-28
fix(FormattingToolbar): fixes change of viewport for link modal - I376

elit-altum:

# Issue #376 

### Changes
- Added the ```window.scroll()``` method to scroll to the HyperlinkModal's viewport
- Also changed the ```HyperLinkWrapper``` and ```HyperLinkCaret``` styles to tab of 2.

### Flags
- Was the indentation of the fields intentionally of 4 tab spaces? If yes, I will revert it!

### DIFF
--- a/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
+++ b/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
@@ -19,40 +19,38 @@ import OpenIcon from '../components/icons/open';
 import DeleteIcon from '../components/icons/delete';
 
 const HyperlinkWrapper = styled.div`
-    position: absolute;
-    z-index: 3000;
-    top: -10000px;
-    left: -10000px;
-    margin-top: -6px;
-    opacity: 0;
-    background-color: #FFFFFF;
-    border: 1px solid #d4d4d5;
-    border-radius: .3rem;
-    transition: opacity 0.75s;
+  position: absolute;
+  z-index: 3000;
+  margin-top: -6px;
+  opacity: 0;
+  background-color: #FFFFFF;
+  border: 1px solid #d4d4d5;
+  border-radius: .3rem;
+  transition: opacity 0.75s;
 
-    min-width: min-content;
-    line-height: 1.4285em;
-    max-width: 250px;
-    padding: .833em 1em;
-    font-weight: 400;
-    font-style: normal;
-    color: rgba(0,0,0,.87);
-    box-shadow: 0 2px 4px 0 rgba(34,36,38,.12), 0 2px 10px 0 rgba(34,36,38,.15);
-    & > * {
-        display: inline-block;
-    }
+  min-width: min-content;
+  line-height: 1.4285em;
+  max-width: 250px;
+  padding: .833em 1em;
+  font-weight: 400;
+  font-style: normal;
+  color: rgba(0,0,0,.87);
+  box-shadow: 0 2px 4px 0 rgba(34,36,38,.12), 0 2px 10px 0 rgba(34,36,38,.15);
+  & > * {
+      display: inline-block;
+  }
 `;
 const HyperlinkCaret = styled.div`
-    position: absolute;
-    z-index: 4000;
-    left: calc(50% - 5px);
-    top: -10px;
-    height: 0;
-    width: 0;
-    border-left: 5px solid transparent;
-    border-right: 5px solid transparent;
-    border-bottom: 10px solid #d4d4d5;
-    transition: opacity 0.75s;
+  position: absolute;
+  z-index: 4000;
+  left: calc(50% - 5px);
+  top: -10px;
+  height: 0;
+  width: 0;
+  border-left: 5px solid transparent;
+  border-right: 5px solid transparent;
+  border-bottom: 10px solid #d4d4d5;
+  transition: opacity 0.75s;
 `;
 
 const LinkIconHolder = styled.div`
--- a/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
+++ b/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
@@ -21,6 +21,8 @@ import DeleteIcon from '../components/icons/delete';
 const HyperlinkWrapper = styled.div`
   position: absolute;
   z-index: 3000;
+  top: -10000px;
+  left: -10000px;
   margin-top: -6px;
   opacity: 0;
   background-color: #FFFFFF;
--- a/src/RichTextEditor/FormattingToolbar/index.js
+++ b/src/RichTextEditor/FormattingToolbar/index.js
@@ -49,7 +49,7 @@ const FormattingToolbar = ({ canBeFormatted, showLinkModal, setShowLinkModal }) 
           + rect.width / 2}px`;
 
       // adjust window scroll to bring focus to the modal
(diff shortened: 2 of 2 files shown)


elit-altum:
@irmerk @DianaLease Please review!

irmerk:
A couple issues I spot on first pass:
1. The modal (and it's caret) are off center
2. The width is off

(1)
---

Screen Shot 2020-04-28 at 8 58 25 AM

(2)
---

Screen Shot 2020-04-28 at 8 58 49 AM

elit-altum:
@irmerk Sorry I didn't see that, my bad. Just fixed 1.
Can you tell me how to recreate the second instance ? I am not able to do that.

irmerk:
Ah the second one isn't an issue anymore. A new issue I'm finding where the screen shoots down when clicking a link:

Apr-28-2020 10-39-20

elit-altum:
> 
> 
> Ah the second one isn't an issue anymore. A new issue I'm finding where the screen shoots down when clicking a link:
> 
> Apr-28-2020 10-39-20

@irmerk If you press Enter near the end of the page on any text. The page scrolls down to that viewport and brings it to center of the screen is that expected behavior ?

elit-altum:
@irmerk One more thing, right now the z-index is such that the link modal is on top of the formatting toolbar icons. Is that expected behavior or should I fix it?

irmerk:
@Michael-Grover do you have any input on the last two comments from @elit-altum?

Michael-Grover:
It looks like on Google docs, the screen never moves when the link modal pops up, even if the link is near the bottom of the page. I think we should use that behavior, so that the screen never unexpectedly moves

elit-altum:
@Michael-Grover This can be implemented but I am running into some issues. 
1) The value of ```window.scrollY``` is always coming out to be as much as the end of the page if I use it the the ```useEffect()``` hook which displays the modal.
2) Outside the hook I tried using state to store the ```window.scrollY``` value and scroll to that after the modal opens. But if I scroll the page with the opened modal, the editor crashes with an error *tried to update state of an unmounted component* is there a workaround for any of these problems ?

irmerk:
> 1. The value of `window.scrollY` is always coming out to be as much as the end of the page if I use it the the `useEffect()` hook which displays the modal.

Hmm, so the `window.scrollY` logic is to keep it from going to the top of the page when a link is clicked? Is it possible to prevent the scroll to the top of the page without doing this? Or maybe only do this if the modal is calculated to be _outside_ of the current window view?

elit-altum:
@irmerk This is possible from the second logic of using the state to hold the ```window.scrollY```. It sounds weird but is there anyway I can just use the value of state once and then not update it later ?

elit-altum:
> @irmerk If you press Enter near the end of the page on any text. The page scrolls down to that viewport and brings it to center of the screen is that expected behavior ?

Can you please tell me about this @irmerk so I can see if this is related?

irmerk:
> @irmerk This is possible from the second logic of using the state to hold the `window.scrollY`. It sounds weird but is there anyway I can just use the value of state once and then not update it later ?

Hmm well I think you can use `useState` and not use the second argument returned:

```js
const [wscroll, setWscroll] = useState(...window.scrollY)
```
If you never change this with `setWscroll`, it should stay the same? Though I might not be fully understanding.

elit-altum:
> 
> 
> > @irmerk This is possible from the second logic of using the state to hold the `window.scrollY`. It sounds weird but is there anyway I can just use the value of state once and then not update it later ?
> 
> Hmm well I think you can use `useState` and not use the second argument returned:
> 
> ```js
> const [wscroll, setWscroll] = useState(...window.scrollY)
> ```
> 
> If you never change this with `setWscroll`, it should stay the same? Though I might not be fully understanding.

@irmerk this does not seem to work as window.scrollY will be used for the initial state setting only. So even if the user does scroll down, the value is still what it was before.

Michael-Grover:
Can someone clarify what functionality is trying to be achieved with the scroll stuff?

elit-altum:
@Michael-Grover When the link modal opens, the screen is scrolling to the top of the page. This could be due to the modal positioning and body height. But what we are trying to achieve is to manually scroll to a particular point on the page and overriding the scroll to top bug.
Hope this helps!

Michael-Grover:
> It looks like on Google docs, the screen never moves when the link modal pops up, even if the link is near the bottom of the page. I think we should use that behavior, so that the screen never unexpectedly moves

I don't think we should move the screen when opening a link. Google docs already has a nice way of handling link modals, I think we should copy that.

Michael-Grover:
When the screen moves unexpectedly it can be disorienting and distracting for users

elit-altum:
@Michael-Grover I have been trying to recreate that only!

Michael-Grover:
sorry I misunderstood. Sounds good

irmerk:
Unsure what else to suggest on this one. @DianaLease any thoughts?

DianaLease:
Played around with this locally and it seems that the page jumping issue is stemming from us setting the focus to be on the URL field when the modal opens. I'm wondering if we should just remove that functionality for now @Michael-Grover ? This page jumping bug seems like higher priority.

Michael-Grover:
@DianaLease I think it's important that people can create a link without using their mouse to select the link input. Maybe I'm wrong but this requires us to set the focus on the URL field when the modal opens, right? Many people like using keyboard shortcuts to make a link, and when products mess with the basic steps for creating link it can be annoying.

1. Write text that I want to turn into a link
2. press alt+shift+backspace to highlight text
3. press cmd+k to make link
4. Type URL
5. Press enter to apply url
6. Press right arrow to deselect highlighted text and move text cursor back to the end of previously highlighted text
7. Continue typing

This sequence of key strokes is identical on Apple Mail, Gmail web client, Google docs, and Microsoft Word (sometimes step 6 is done automatically), and most other editors. To insert an additional step of having to move your mouse to click on the URL field, or having to tab a few times to focus the URL field, would throw a lot of people off. Slack, for example, does not use cmd+K as their hyperlink shortcut and it always trips me up when I try to make a link.

DianaLease:
@Michael-Grover You'd still be able to add a URL with your keyboard. You'd just have to hit tab after `cmd+k`

Michael-Grover:
I still think that's very disruptive to peoples' flow when they are used to being able to add a link in one second from muscle memory. Would people necessarily know that clicking tab will focus the correct field? I think they may just use their mouse rather than make an assumption of where focus will move when pressing tab.

Is the main problem that when the hyperlink field goes off-screen, the screen moves? Can we do what Google does and make the modal appear above the selected text when it is at the bottom of the screen?

image

If there is space for modal below highlighted text, put the modal below
image

If there is not space for modal below highlighted text, put modal above
image

If the highlighted text is off screen, move screen to text such that the text is in the middle of the screen and place modal above
image

DianaLease:
@Michael-Grover I think I found a solution that allows us to keep the focus and not have the page jump. @elit-altum I'll put in a PR and tag you to check it out if that's alright

irmerk:
It'll be great if @DianaLease's PR works, but I think for context this is two separate issues. @Michael-Grover I agree that that is the ideal outcome, but we want to keep these PRs and Issues as narrow focused as possible. So if this would solve 1/2 problems, we shouldn't block it being merged and can follow up on the remaining issue with a new GitHub Issue.

DianaLease:
PR here: https://github.com/accordproject/markdown-editor/pull/388

PR #384, opened 2020-04-29
fix(toolbarHelpers): update type of node  instead of wrapping node if updating heading style

DianaLease:

# Issue 
- Previously when adding a heading, we would wrap the children nodes in a heading node, resulting in a paragraph inside of a heading

### Changes
- Update type of node using `setNodes` instead of wrapping node if going from paragraph to heading or vice versa.

## FLAGS
- Please test links in headings

### DIFF
--- a/src/RichTextEditor/utilities/toolbarHelpers.js
+++ b/src/RichTextEditor/utilities/toolbarHelpers.js
@@ -27,6 +27,11 @@ export const toggleBlock = (editor, format) => {
   Transforms.unwrapNodes(editor, { match: n => isListItem(n.type), split: true });
   Transforms.unwrapNodes(editor, { match: n => LIST_TYPES.includes(n.type), split: true });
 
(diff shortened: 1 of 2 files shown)


src/RichTextEditor/utilities/toolbarHelpers.js
```diff
@@ -27,6 +27,11 @@ export const toggleBlock = (editor, format) => {
   Transforms.unwrapNodes(editor, { match: n => isListItem(n.type), split: true });
   Transforms.unwrapNodes(editor, { match: n => LIST_TYPES.includes(n.type), split: true });
 
+  if (format === 'paragraph' || format.startsWith('heading')) {
```
jeromesimeon, line 30: This will allow to turn paragraph into heading and vice versa, is that correct?

DianaLease: correct!

Issue #365, opened 2020-04-18
Hyperlink Modal accepts invalid link values


nik72619c:
**Describe the bug**
The hyperlink modal accepts invalid link URLs. For eg:- a space, normal strings, etc.

**To Reproduce**
1. select a text
2. Click on the `Link` button on the top to convert the text to a link
3. enter an invalid link
4. The text converts to a link which points to the invalid link URL

**Expected behavior**
An error or at least a warning message should be displayed, to alert the user that the link mentioned is invalid

**Desktop (please complete the following information):**
 - OS: Mac OS
 - Browser Chrome
 - Version 80.0.3987.163 (Official Build) (64-bit)

nik72619c:
@DianaLease @dselman I feel we can integrate `react-form` into this to apply validations. Would love to hear your thoughts on this :)

irmerk:
I think Google Docs allows this. @Michael-Grover?

DianaLease:
> @DianaLease @dselman I feel we can integrate `react-form` into this to apply validations. Would love to hear your thoughts on this :)

If we add a library, I'm a fan of using https://github.com/final-form/react-final-form as it is well-maintained and lightweight while still offering a lot!

Michael-Grover:
I believe in another issue there was a decision that this error validation isn't necessary but I can't find it

nik72619c:
I am getting this error, after installing `react-final-form`, can someone help?

<img width="935" alt="Screenshot 2020-04-21 at 9 02 44 PM" src="https://user-images.githubusercontent.com/30630904/79885107-df128780-8413-11ea-969a-e51057dfd41d.png">

nik72619c:
Fixed.
Also, I wanted to ask if we would require a validation library such as `react-final-form` in this case?
There are not many validations required since the repo is not using forms extensively?

Michael-Grover:
I don't think this is necessary. Google Docs doesn't do this kind of validation. Are there examples of this kind of hyperlink validation in any prominent apps?

irmerk:
Yeah I don't think this needs to be that thorough. If the URL does not begin with `http://` or `https://`, add it. That should probably be the extent we do at this point.

nik72619c:
> Yeah I don't think this needs to be that thorough. If the URL does not begin with `http://` or `https://`, add it. That should probably be the extent we do at this point.

@irmerk shouldn't the logic be that if the URL begins with `http://` or `https://`, add it ?

[...]

PR #378, opened 2020-04-24
feat(hyperlink modal): fixed invalid url acceptance issue

nik72619c:
Signed-off-by: Nikhil <nikhilsharmarockstar21@gmail.com>

# Issue #365 
This PR adds the feature to validate whether the URL contains `http://` or `https://` in it.

### Changes
- Added logic for validating URL and maintained a state variable to show an error message
- Changed the placeholder from `https://example.com` to `example.com`

PR #386, opened 2020-04-30
feat(hyperlink Modal): fix hyperlink modal accepts invalid urls

nik72619c:
Signed-off-by: Nikhil <nikhilsharmarockstar21@gmail.com>

# Issue #365 
This PR adds the feature to insert `https://` if the URL does not start with a valid prefix

### Changes
- Added logic for prefixing the URL with `https://` if the URL does not start with `http://` or `https://`

### DIFF
--- a/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
+++ b/src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
@@ -161,9 +161,18 @@ const HyperlinkModal = React.forwardRef(({ ...props }, ref) => {
     props.setShowLinkModal(false);
   };
 
+  const validateUrl = (url) => {
+    const isUrlInvalid = !(url.startsWith('http://') || url.startsWith('https://'));
+    if (isUrlInvalid) {
+      url = `https://${url}`;
+    }
+    return url;
+  };
+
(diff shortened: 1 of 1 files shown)


nik72619c:
@irmerk @DianaLease @Michael-Grover ☝️ 
Ref #378

src/RichTextEditor/FormattingToolbar/HyperlinkModal.js
```diff
@@ -161,9 +161,18 @@ const HyperlinkModal = React.forwardRef(({ ...props }, ref) => {
     props.setShowLinkModal(false);
   };
 
+  const validateUrl = (url) => {
+    const isUrlInvalid = !(url.startsWith('http://') || url.startsWith('https://'));
+    if (isUrlInvalid) {
+      url = `https://${url}`;
+    }
+    return url;
+  };
```
irmerk, line 170: I think this could be written:

```js
    if (isUrlInvalid) { return `https://${url}`; }
    return url;
```

nik72619c: @irmerk makes sense. Made the required changes :)

Issue #379, opened 2020-04-24
Display a message in hyperlink modal to indicate that the link is copied


nik72619c:
**Is your feature request related to a problem? Please describe.**
Currently, when a user clicks on the copy button to copy the link to the clipboard, there is no message which tells the user that the link is copied.

**Describe the solution you'd like**
A simple message could be displayed in the modal indicating that the link is copied

PR #391, opened 2020-05-01
feat(hyperlink modal): added message for successful copy of a link url

nik72619c:
Signed-off-by: Nikhil <nikhilsharmarockstar21@gmail.com>

# Issue #379 

### Changes
- Created a new styled-component for displaying the success message
- added state to toggle the message on copying the icon for the same


nik72619c:
@DianaLease @irmerk @Michael-Grover please have a look :)

Michael-Grover:
I'm not sure if this is entirely necessary. Google Docs doesn't do this but that's the only example I have had a chance to check. Do you know of any examples of a hyperlink copy confirmation in popular tools so we can see how they handle it?

If we choose to do this, I think the confirmation message should appear to the right of the icons so that the size of the modal and the position of the icons doesn't abruptly change
image

nik72619c:
thanks for the review @Michael-Grover . Making the changes !

elit-altum:
@nik72619c If I copy a link once. The message will show. If I then change the link to some other value, and then try to copy it again in the same modal,the message still stays up and there would be no way for me to know that the new link was copied correctly or not. 

Should we make ```linkCopied``` to false when the link value changes?
What do you say?

irmerk:
We're migrating this repository into [`web-components`](https://github.com/accordproject/web-components/) the next couple of days, so I'll be closing this soon and you should reopen there.

PR #393, opened 2020-05-06
refactor(*): refactor & clean up

DianaLease:

- Refactors to fix a bunch of console warnings
-  replaces `div`s that wrap image component with `span` so that images can be within in paragraphs in the tree
- Updates semantic-ui-react
- FYI - I deleted a bunch of files used by `src/RichTextEditor/components/index.js` and instead defined those elements inline because it was only one line of code for each. And if we were to keep those as separate components, a lot of them would need their `ref` forwarded, so I thought it was cleaner this way.


src/RichTextEditor/FormattingToolbar/ToolbarMenu.js
```diff
@@ -25,4 +25,6 @@ const ToolbarMenu = React.forwardRef(
   ({ ...props }, ref) => (<Menu {...props} ref={ref} />)
 );
 
+ToolbarMenu.displayName = 'ToolbarMenu';
```
irmerk, line 28: Can you remove the `// eslint-disable-next-line react/display-name` from above here too?

DianaLease: Yes, good catch!

PR #396, opened 2020-05-08
feat(RichTextEditor): allow consumer to prevent actions

irmerk:
# No Issue
Allow consumer of `markdown-editor` to prevent certain actions

### Changes
- Sets groundwork for preventing `onKeyDown`
- Sets groundwork for preventing `onCopy` and `onCut`

### Flags
- This will be done in conjunction with work in `cicero-ui` where the Contract Editor can prevent <kbd>BACKSPACE</kbd> when cursor is on the last node but the previous node is a smart clause template.
- This will also be done in conjunction with work in `cicero-ui` where the Contract Editor can prevent <kbd>CMD</kbd>+<kbd>C</kbd> and <kbd>CMD</kbd>+<kbd>X</kbd> when the selection begins or ends inside a smart clause template.

### Related Issues
- N/A

### DIFF
--- a/src/RichTextEditor/index.js
+++ b/src/RichTextEditor/index.js
@@ -23,7 +23,13 @@ import { withLists } from './plugins/withLists';
 import FormatBar from './FormattingToolbar';
 
 const RichTextEditor = (props) => {
-  const { augmentEditor, isEditable, canBeFormatted } = props;
+  const {
+    canCopy,
+    canKeyDown,
+    augmentEditor,
+    isEditable,
+    canBeFormatted
+  } = props;
   const [showLinkModal, setShowLinkModal] = useState(false);
   const editor = useMemo(() => {
     if (augmentEditor) {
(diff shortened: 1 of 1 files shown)


src/RichTextEditor/index.js
```diff
@@ -90,6 +99,7 @@ const RichTextEditor = (props) => {
 
   const handleCopyOrCut = useCallback((event, cut) => {
     event.preventDefault();
+    if (canCopy && !canCopy(editor)) return;
```
DianaLease, line 102: If you want to avoid having to check for `canCopy` before calling it, you can pass a default prop for it that is a function that returns `true` like we do with `canBeFormatted` on line 189. Then if a function isn't passed, that default one will be called and will always return true (ie. it can be copied and we won't hit the return on line 102).

irmerk: So in the deconstruct of `props`?

```js
const RichTextEditor = (props) => {
  const {
    canCopy || () => true,
    canKeyDown,
    augmentEditor,
    isEditable,
    canBeFormatted
  } = props;
```

DianaLease: No, in the `defaultProps` object at the end of the file where we have:
```
RichTextEditor.defaultProps = {
  isEditable: () => true,
  canBeFormatted: () => true,
};
```

